PR #8, opened 2020-05-08
[lua-monitoring] Find out ip address information of each interface #1

### LINKED ISSUE #1: [lua-monitoring] Find out ip address information of each interface
This issue depends on https://github.com/openwisp/openwisp-monitoring/issues/30.

In this script:
https://github.com/openwisp/lua-monitoring/blob/master/netjson-monitoring.lua

For each monitored interface, we have to find out the ip addresses.
Once we get the list of IP, for each ip we can check the configuration of that interface to see if any static address is configured, and find out if the IP we got matches the static IP configured, if yes, that's a static address and we'll flag it as such, otherwise it's most probably an address got in DHCP client mode or manually configured, we can further check the configuration to see if the interface is flagged as proto dhcp, if yes we'll flag the address as dhcp, otherwise we'll flag it as static.

nepython:
Closes #1

### DIFF
--- a/netjson-monitoring.lua
+++ b/netjson-monitoring.lua
@@ -155,6 +155,39 @@ function is_excluded(name)
   end
 end
 
+-- collect interface addresses
+function get_addresses(name)
+    interface_data = ubus:call('network.interface', 'dump', {})
+    addresses = {}
+    interface_list = interface_data['interface']
+    for _, interface in pairs(interface_list) do
+        if interface['l3_device'] == name then
+            if interface['ipv4-address'] and next(interface['ipv4-address'])then
+                family = 'ipv4'
+            elseif interface['ipv6-address'] and next(interface['ipv6-address'])then
+                family = 'ipv6'
+            end
+            if interface[family..'-address'] then
+                for _,address in pairs(interface[family..'-address']) do
+                    if address['proto'] == 'static' then
+                        proto = 'static'
+                    else
+                        -- expected values are: dhcp, dhcpv6
+                        proto = 'dhcp'
+                    table.insert(addresses, {
+                        address = address['address'],
+                        mask = address['mask'],
+                        proto = proto,
+                        family = family
+                    })
+                    end
+                end
+            end
+        end
+    end
+    return addresses
+end
+
 -- collect device data
 network_status = ubus:call('network.device', 'status', {})
 wireless_status = ubus:call('network.wireless', 'status', {})
@@ -182,7 +215,8 @@ for radio_name, radio in pairs(wireless_status) do
                     signal = iwinfo.signal,
                     noise = iwinfo.noise,
                     country = iwinfo.country
-                }
+                },
+                addresses = get_addresses(name)
             }
             if clients and next(clients.clients) ~= nil then
               netjson_interface.wireless.clients = netjson_clients(clients.clients)


netjson-monitoring.lua
```diff
@@ -72,6 +72,24 @@ netjson = {
     }
 }
 
+-- collect interface addresses
+interface_data = ubus:call('network.interface', 'dump', {})
+addresses = {}
```
PabloCastellano, line 77: Please move these 2 lines inside get_address()

netjson-monitoring.lua
```diff
@@ -72,6 +72,24 @@ netjson = {
...
+
+function get_address()
```
PabloCastellano, line 79: To be coherent, can you rename it to get_addresses()?

netjson-monitoring.lua
```diff
@@ -72,6 +72,24 @@ netjson = {
...
+  for _, interface_list in pairs(interface_data) do
```
PabloCastellano, line 80: Since an array with the interface key is expected, let's do `interface_list = interface_data['interface']`

netjson-monitoring.lua
```diff
@@ -72,6 +72,24 @@ netjson = {
...
+    for _, interface in pairs(interface_list) do
+      if interface["ipv4-address"] then
+        table.insert(addresses, interface["ipv4-address"][1])
```
PabloCastellano, line 83: `interface["ipv4-address"]` is a list, so you need to loop over the it because there might be multiple ips associated to the interface

netjson-monitoring.lua
```diff
@@ -72,6 +72,31 @@ netjson = {
...
+                end
+            elseif interface['ipv6-address'] then
+                for _,address in pairs(interface['ipv6-address']) do
+                    table.insert(addresses, address)
+                end
+            end
+            if interface.proto == 'static' then
+                addresses[#addresses].proto = 'static'
+            else
+                addresses[#addresses].proto = 'dhcp'
```
PabloCastellano, line 94: At the moment `proto` looks like it's only added for the last element in the address array.

I propose to use the following:

```
        if interface['l3_device'] == name then
            if interface['ipv4-address'] then
                for _,address in pairs(interface['ipv4-address']) do
                    if address['proto'] == 'static' then
                        proto = 'static'
                    else
                        -- expected values are: dhcp, dhcpv6
                        proto = 'dhcp'
                    table.insert(addresses, {
                        address = address['address'],
                        mask = address['mask'],
                        proto = proto,
                        family = 'ipv4'
                    })
                end
            elseif interface['ipv6-address'] then
                .. similar for ipv6
```

netjson-monitoring.lua
```diff
@@ -72,6 +72,39 @@ netjson = {
...
+                    if address['proto'] == 'static' then
+                        proto = 'static'
+                    else
+                        -- expected values are: dhcp, dhcpv6
+                        proto = 'dhcp'
+                    table.insert(addresses, {
+                        address = address['address'],
+                        mask = address['mask'],
+                        proto = proto,
+                        family = family
```
PabloCastellano, line 98: you must send either `ipv4` or `ipv6`, not `ipv#-address`!

nepython: My bad

netjson-monitoring.lua
```diff
@@ -72,6 +72,39 @@ netjson = {
...
+    addresses = {}
+    interface_list = interface_data['interface']
+    for _, interface in pairs(interface_list) do
+        if interface['l3_device'] == name then
+            if interface['ipv4-address'] and next(interface['ipv4-address'])then
+                family = 'ipv4'
+            elseif interface['ipv6-address'] and next(interface['ipv6-address'])then
+                family = 'ipv6'
+            end
+            if interface[family] then
```
PabloCastellano, line 87: This will not work because the key is 'ipv#-address'

nemesisdesign:
@nepython https://github.com/openwisp/lua-monitoring/commit/7bbab6e57baec6702c11085464cbc0d4fa779cfa

PR #11, opened 2020-05-15
[enhancement] Add disk usage to resources

nepython:
Required for `openwisp-monitoing #43`.

I had to refactor the **print(cjon.encode(..))** part primarily because it converted `/` to `\/` (expected behavior but unwanted by us). For e.g. `name: \/dev\/root` would be output hence made slight change there. Also, I first stored it in a variable and didn't output it directly because `gsub` also returns instances of replacement (again not required)


netjson-monitoring.lua
```diff
@@ -253,4 +275,6 @@ if next(interfaces) ~= nil then
     netjson.interfaces = interfaces
 end
 
-print(cjson.encode(netjson))
+-- json.encode converts `/` to `\/` hence the workaround
+devicedata = cjson.encode(netjson):gsub('\\', '')
```
PabloCastellano, line 279: Can you provide an example of why this is necessary? My '/' aren't being converted to '\/'

netjson-monitoring.lua
```diff
@@ -119,6 +119,27 @@ load_average = {tonumber(loadavg_output[1]),
...
+function parse_disk_usage()
+    file = io.popen('df')
+    disk_usage_info = {}
+    for line in file:lines() do
+        if line:sub(1, 10) ~= 'Filesystem' then
+            filesystem, memory, used, available, percent, location = line:match("(%S+)%s+(%S+)%s+(%S+)%s+(%S+)%s+(%S+)%s+(%S+)")
+            table.insert(disk_usage_info, {
+                name = filesystem,
+                -- in 1k blocks
+                size = memory,
```
PabloCastellano, line 131: let's rename for consistency memory to size too (size = size)

netjson-monitoring.lua
```diff
@@ -119,6 +119,27 @@ load_average = {tonumber(loadavg_output[1]),
...
+                used_space = used,
+                available_space = available,
+                percent_used = percent,
```
PabloCastellano, line 134: Can you remove the trailing % sign?

netjson-monitoring.lua
```diff
@@ -119,6 +119,27 @@ load_average = {tonumber(loadavg_output[1]),
...
+    file = io.popen('df')
+    disk_usage_info = {}
+    for line in file:lines() do
+        if line:sub(1, 10) ~= 'Filesystem' then
+            filesystem, memory, used, available, percent, location = line:match("(%S+)%s+(%S+)%s+(%S+)%s+(%S+)%s+(%S+)%s+(%S+)")
+            table.insert(disk_usage_info, {
+                name = filesystem,
+                -- in 1k blocks
+                size = memory,
+                used_space = used,
```
PabloCastellano, line 132: It would be nice to send these data as numbers. You can use tonumber(used). I have quickly tried it but it failed so you will need to investigate why :wink:

netjson-monitoring.lua
nepython: **Without it**: `"name":"\/dev\/root","mounted_on":"\/"`

**Expected**: `"name":"/dev/root","mounted_on":"/"`

I mentioned it above in the issue opening comment :sweat_smile:

netjson-monitoring.lua
nepython: Sure, took care of it on the openwisp instance side but if you say I will make those changes here

netjson-monitoring.lua
```diff
@@ -119,6 +119,27 @@ load_average = {tonumber(loadavg_output[1]),
...
                 tonumber(loadavg_output[3])}
 
+function parse_disk_usage()
+    file = io.popen('df')
+    disk_usage_info = {}
+    for line in file:lines() do
+        if line:sub(1, 10) ~= 'Filesystem' then
+            filesystem, memory, used, available, percent, location = line:match("(%S+)%s+(%S+)%s+(%S+)%s+(%S+)%s+(%S+)%s+(%S+)")
+            table.insert(disk_usage_info, {
+                name = filesystem,
```
PabloCastellano, line 129: Are the key elements of the array defined somewhere already?

Otherwise I would propose using these more descriptive ones: 
* filesystem
* size_bytes
* used_bytes
* available_bytes
* used_percent
* mountpoint

@nemesisdesign what do you think?

PabloCastellano: Also I'm unsure about sending used_percent, since it can be calculated.

netjson-monitoring.lua
PabloCastellano: That is very interesting, I'm seeing `/dev/root` in both cases. We need a third person to test it :)

In any case, can you just apply this `gsub('\\', '') to name and mounted_on, instead of the whole json output?

nepython: I noticed this now, will do it

nepython: @PabloCastellano, I mean that the `name` and `mounted_on` are both appearing properly if you print them. But this happens (atleast on my VM) when encoding the table to a json. that is `\/` occurs when you do `cjson.encode(netjson)` and before that everything is fine

PabloCastellano: Yeah in my case I don't see any problem running it on a real device. Are you connected to the VM using SSH? It might be a terminal encoding issue, dunno.

nepython: Yes, I am connected to the VM using SSH. In anycase if someone else doesn't find this issue it can be easily removed (2 lijnes) and it shouldn't be changing anything else since `\` is an escape character which is normally unexpected

PabloCastellano: @nemesisdesign Can you try it and share your output?

netjson-monitoring.lua
PabloCastellano: @nepython Remember to change all parameters to the more descriptives names I mentioned above. We decided to send all the data at the moment, including `used_percent`

netjson-monitoring.lua
PabloCastellano: Since pandafy confirmed that he doesn't have this problem neither in a VM, I assume there is an encoding issue somewhere in your setup. In any case, the recommended way of developing for OpenWISP is using a real device.

netjson-monitoring.lua
```diff
@@ -130,7 +152,8 @@ netjson = {
     resources = {
         load = load_average,
         memory = system_info.memory,
-        swap = system_info.swap
+        swap = system_info.swap,
+        disk_usage = parse_disk_usage()
```
PabloCastellano, line 156: As @nemesisdesign mentioned, it would be very useful to also send the number of cores available in the system so that we can calculate the % of cpu use later.

Do you mind adding it in this PR too?

You can easily get it from the output of `cat /proc/cpuinfo | grep -c processor`

PabloCastellano, line 156: For coherence, please rename disk_usage to just 'disk'

netjson-monitoring.lua
```diff
@@ -119,6 +119,29 @@ load_average = {tonumber(loadavg_output[1]),
...
+                used_percent = tonumber(percent),
+                mount_point = location,
+        })
+        end
+    end
+    file:close()
+    return disk_usage_info
+end
+
+processors = io.popen('cat /proc/cpuinfo | grep -c processor')
```
PabloCastellano, line 143: For coherency, please create a get_cpus function so that you can then add the data such as `cpus = get_cpus()`

nepython:
Done :smiley:

collect_spec.lua
```diff
@@ -0,0 +1,34 @@
+-- os = require('os')
```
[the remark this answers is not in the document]
nepython: It was a test script so I could test functions individually. Didn't notice it got commited :sweat:.

PR #17, opened 2020-07-15
[feature] Monitor addresses of all interfaces by default #12

### LINKED ISSUE #12: [lua-monitoring] Monitor addresses of all interfaces by default
Monitor addresses of all interfaces by default, but do not include traffic statistics unless explicitly configured. 

I also want to make the monitoring information regarding addresses more complete by using:

```lua
nixio = require('nixio')
cjson = require('cjson')
print(cjson.encode(nixio.getifaddrs()))
```

Because `ubus call network.interface dump` only gives partial information, although it's very useful as well, so I think I will try to merge the two.

Possible new algorithm:

- call `ubus call network.interface dump`, build a dictionary of all the interfaces and their addresses
- call `nixio.getifaddrs()`, update the dictionary created previously and add only new information
- for each interface included in the list of interfaces passed in the first argument to [netjson-monitoring.lua](https://github.com/openwisp/lua-monitoring/blob/master/netjson-monitoring.lua), call also `ubus call network.device status '{"name": "<NAME>"}'` to get additional information and traffic statistics
- prepare the netjson output, including statistic/traffic information only for the interfaces specified by configuration

PabloCastellano:
* Retrieve more interfaces using the nixio library.
    * Detect link local and ULA ipv6 addresses.
    * Add two helpers: has_value and starts_with.
    * Support --stats as the 2nd position parameter to optionally include
      statistics data.
    * Add the following interface data: type, up, mac, txqueuelen, mtu, speed,
      bridge_members, stp, dns-search, dns-server and link-supported.
      NOTE: depending on the device and openwrt version some data might not be
      available
    * Detect default gateway

Closes #12

TODO (I can't do more because my device doesn't provide these data):

* [ ] Extend interface types in netjson definition (vlan, bridge, virtual, other, ethernet, wireless)
* [x] bridge_members
* [x] stp
* [ ] Heuristic to guess device type when type is "Network device"
* [ ] speed (not present in any of my interfaces)
* [ ] link_supported (not present in any of my interfaces)
* [x] route/gateway/nexthop
* [x] dns-server
* [x] dns-search

### DIFF
--- a/netjson-monitoring.lua
+++ b/netjson-monitoring.lua
@@ -4,6 +4,7 @@
 io = require('io')
 ubus_lib = require('ubus')
 cjson = require('cjson')
+nixio = require('nixio')
 uci = require('uci')
 uci_cursor = uci.cursor()
 
@@ -25,6 +26,19 @@ function split(str, pat)
       table.insert(t, cap)
    end
    return t
+end
+
+local function has_value (tab, val)
+    for index, value in ipairs(tab) do
+        if value == val then
+            return true
+        end
+    end
+    return false
+end
+
+local function starts_with(str, start)
+   return str:sub(1, #start) == start
 end
 
 -- parse /proc/net/arp
@@ -110,7 +124,6 @@ function parse_dhcp_lease_file(path, leases)
 end
 
 function get_dhcp_leases()
-    local uci_cursor = uci.cursor()
     local dhcp_configs = uci_cursor:get_all('dhcp')
     local leases = {}
 
@@ -226,6 +239,9 @@ if included_interfaces then
   end
 end
 
+-- determine the interfaces to monitor
+include_traffic_stats = arg[2] and arg[2] == '--stats'
+
 function is_excluded(name)
   if next(included) ~= nil then
     return included[name] == nil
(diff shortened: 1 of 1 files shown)


PabloCastellano:
Thanks! I've made some changes already but still have some issues:

* Regarding interface type "Network Device"
    * I couldn't find any "wifi info" or "link-advertising" data in my device, which has both of them. Do you have any example?
* None of my interfaces have `speed` or `link-supported` attributes
* Regarding interface has an ipv4 starting with 127.0.0 it's loopback
    * At the moment loopback is ignored, check the function `is_excluded`. Should I remove this limitation?

nemesisdesign:
> Thanks! I've made some changes already but still have some issues:
> 
> * Regarding interface type "Network Device"
>   
>   * I couldn't find any "wifi info" or "link-advertising" data in my device, which has both of them. Do you have any example?
> * None of my interfaces have `speed` or `link-supported` attributes
> * Regarding interface has an ipv4 starting with 127.0.0 it's loopback
>   
>   * At the moment loopback is ignored, check the function `is_excluded`. Should I remove this limitation?

You're right, 

> * dns_server and dns_search
>   
>   * According to the netjson definition, these belong to DeviceConfiguration instead of DeviceMonitoring. Should I still add them?

Yes please, because in many cases dns settings are sent by the DHCP server and the network operator would have to SSH into the device via the management VPN to find out. 

> * please also check the config in /etc/config/network (using the value available in "interface" to access the relevant openwrt config) to see if the spanning tree protocol is enabled (stp, defaults to false), please handle errors if the config is not available (defaults to false)
>   
>   * What should I do if stp is enabled? Where in the array should I include this value?

in the interface, right after `bridge_members`.

PabloCastellano:
> > Thanks! I've made some changes already but still have some issues:
> > 
> > * Regarding interface type "Network Device"
> >   
> >   * I couldn't find any "wifi info" or "link-advertising" data in my device, which has both of them. Do you have any example?
> 
> You have to use the information got retrieved:
> https://github.com/openwisp/lua-monitoring/blob/master/netjson-monitoring.lua#L276
> We're basically merging information that is scattered around in a single NetJSON data structure.


Still unclear to me. This is what `ubus call network.wireless status` returns for me:

```
{
	"radio0": {
		"up": false,
		"pending": false,
		"autostart": true,
		"disabled": true,
		"retry_setup_failed": false,
		"config": {
			"channel": "11",
			"hwmode": "11g",
			"path": "platform/10300000.wmac",
			"htmode": "HT20",
			"disabled": true
		},
		"interfaces": [
			
		]
	}
}
```



radio0 doesn't show up in the output of `ubus call network.device status` nor `ubus call network.interface dump`.

> > * None of my interfaces have `speed` or `link-supported` attributes
> 
> It's in `ubus call network.device status`, I'll send you examples via email.



Not in my machine. I'm using openwrt version 19.07.3, what about you?

My outputs are available here: https://gist.github.com/PabloCastellano/c19bd6c6228c0934c2bdeb1a47416d55

PabloCastellano:
@nemesisdesign Installing `wpad` made no difference. I had `rpcd-mod-iwinfo` already.

> I'm also using OpenWRT 19, but I also do not get the same amount of info on different openwrt versions and devices, what matters is that it's used when available.

Unfortunately I can't do anything when type is "Network device" and none of "speed" and "link-supported" are present. Do you think we can use any other field to guess if it's an ethernet or a wireless device?

nemesisdesign:
> @nemesisdesign Installing `wpad` made no difference. I had `rpcd-mod-iwinfo` already.
> 
> > I'm also using OpenWRT 19, but I also do not get the same amount of info on different openwrt versions and devices, what matters is that it's used when available.
> 
> Unfortunately I can't do anything when type is "Network device" and none of "speed" and "link-supported" are present. Do you think we can use any other field to guess if it's an ethernet or a wireless device?

I can't think of anything reliable at the moment, just set it to `other`

netjson-monitoring.lua
```diff
@@ -236,45 +253,108 @@ function is_excluded(name)
...
-                    address = address['address'],
-                    mask = address['mask'],
-                    proto = proto,
+                elseif family == 'inet6' then
                     family = 'ipv6'
-                })
+                    if starts_with(addr, 'fe80') then
+                        proto = 'static'
+                    else
+                        ula = uci_cursor.get("network", "globals", "ula_prefix")
```
nemesisdesign, line 333: be consistent with the quoting style used

netjson-monitoring.lua
```diff
@@ -320,8 +399,28 @@ for name, interface in pairs(network_status) do
...
+        up = interface.up,
+        mac = interface.macaddr,
+        txqueuelen = interface.txqueuelen,
+        mtu = interface.mtu,
+        speed = interface.speed,  -- XXX: untested
+        bridge_members = interface['bridge-members'],
+        stp = interface.stp,  -- XXX: untested
     }
+    if interface.type == 'Network device' then
+        if interface['link-supported'] then
```
nemesisdesign, line 412: we should check that this is not empty (it could be an empty table) because otherwise also openvpn tap interfaces are flagged as ethernet

netjson-monitoring.lua
```diff
@@ -236,45 +252,108 @@ function is_excluded(name)
...
+        mask = address['mask'],
+        proto = proto,
+        family = family,
+        gateway = find_default_gateway(interface.route),
+    }
+    if next(interface['dns-search']) then
+        new_address.dns_search = interface['dns-search']
+    end
+    if next(interface['dns-server']) then
+        new_address.dns_server = interface['dns-server']
```
nemesisdesign, line 287: DNS info should be added to the main netjson object, not the interface object

PabloCastellano: @nemesisdesign It should be in the interface object rather than in the adresses object, right? Otherwise I don't understand it

nemesisdesign: > @nemesisdesign It should be in the interface object rather than in the adresses object, right? I changed it to this way. Otherwise I don't understand it

Not on the interface but in the main `netjson` table:

```lua
netjson.dns_search = dns_info.search
netjson.dns_servers = dns_info.servers  
```

Should be `servers` rather than `server`.

netjson-monitoring.lua
```diff
@@ -320,12 +415,40 @@ for name, interface in pairs(network_status) do
...
+        txqueuelen = interface.txqueuelen,
+        mtu = interface.mtu,
+        speed = interface.speed,  -- XXX: untested
+        bridge_members = interface['bridge-members'],
+        stp = interface.stp,  -- XXX: untested
     }
+    if interface.type == 'Network device' then
+        link_supported = interface['link-supported']
+        if link_supported and next(link_supported) then
+            -- XXX: untested
```
nemesisdesign, line 430: we can remove this, I tested it on my device and it works, we'll keep improving it upon further testing in the future

netjson-monitoring.lua
```diff
@@ -320,12 +415,40 @@ for name, interface in pairs(network_status) do
...
+            netjson_interface.type = 'ethernet'
+            netjson_interface.link_supported = link_supported
+        else
+            netjson_interface.type = 'other'
+            -- TODO: guess 'virtual'
```
nemesisdesign, line 435: maybe we can remove this comment as well

netjson-monitoring.lua
```diff
@@ -236,45 +252,123 @@ function is_excluded(name)
...
+  return nil
+end
+
+-- collect device data
+network_status = ubus:call('network.device', 'status', {})
+wireless_status = ubus:call('network.wireless', 'status', {})
+interface_data = ubus:call('network.interface', 'dump', {})
+nixio_data = nixio.getifaddrs()
+
+
```
nemesisdesign, line 270: there's 2 blank lines but it seems to me we're using 1 blank line almost everywhere in this file

PR #24, opened 2021-01-19
[change] Invert rx & tx bytes on interfaces that are a part of a bridge

### LINKED ISSUE #22: [change] Invert upload/download on interfaces that are member of a bridge
In Linux, bridged interface show tx/rx packets in reverse: download is upload and upload is download.

I couldn't find a canonical reference on why this is the way it is, if anyone finds it, let me know.

Now the problem we have is that in the traffic graph, the upload is greater than the download on wifi interfaces that are part of a bridge, which most users will consider as a bug.

Screenshot from 2021-01-10 18-01-07

So I think we should do the following:

- whenever an interface is flagged as an interface which should be traffic monitored, we should check whether this interface is part of a bridge
- if it is, we should simply invert rx and tx

**Addendum**: I'm trying to do some more research on the subject before proceeding with this: https://forum.openwrt.org/t/inverted-tx-rx-packets-in-bridged-interfaces/85093.

R9295:
closes #22

### DIFF
--- a/netjson-monitoring.lua
+++ b/netjson-monitoring.lua
@@ -460,16 +460,16 @@ end
 
 
 function is_part_of_bridge(name)
-	for _, interface in pairs(network_status) do
-    	if interface['bridge-members'] then
-        	for _, key in pairs(interface['bridge-members']) do
-            	if key  == name then
-					return true
+    for _, interface in pairs(network_status) do
+        if interface['bridge-members'] then
+            for _, key in pairs(interface['bridge-members']) do
+                if key  == name then
+                    return true
                 end
             end
         end
     end
-	return false
+    return false
 end
 
 
@@ -516,12 +516,12 @@ for name, interface in pairs(network_status) do
             end
         end
         if include_stats[name] then
-			if is_part_of_bridge(name) then
+            if is_part_of_bridge(name) then
                 --- Inversing the rx and tx values here as members of a bridge
                 --- have those values inverted by default
                 interface.statistics = inverse_rx_and_tx(interface.statistics)
-			end
-			netjson_interface.statistics = interface.statistics
+            end
+            netjson_interface.statistics = interface.statistics
         end
         addresses = get_addresses(name)
         if next(addresses) then
--- a/netjson-monitoring.lua
+++ b/netjson-monitoring.lua
@@ -458,7 +458,6 @@ for radio_name, radio in pairs(wireless_status) do
     end
 end
 
-
 function is_part_of_bridge(name)
     for _, interface in pairs(network_status) do
         if interface['bridge-members'] then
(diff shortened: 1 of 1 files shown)


netjson-monitoring.lua
```diff
@@ -458,6 +458,20 @@ for radio_name, radio in pairs(wireless_status) do
...
+      if interface['bridge-members'] then
+        for _, key in pairs(interface['bridge-members']) do
+          if key  == name then
+            return true
+          end
+        end
+      end
+    end
+    return false
+end
```
nemesisdesign, line 473: indentation is inconsistent, the first block is indented with 4 spaces but should be 2

nemesisdesign (changes requested):
looks good enough for now, apart from the issue mentioned below, I will test it in the next days

netjson-monitoring.lua
```diff
@@ -489,7 +503,12 @@ for name, interface in pairs(network_status) do
             end
         end
         if include_stats[name] then
-            netjson_interface.statistics = interface.statistics
+			if is_part_of_bridge(name) then
+				local tx_bytes = interface.statistics.tx_bytes
+				interface.statistics.tx_bytes = interface.statistics.rx_bytes
+				interface.statistics.rx_bytes = tx_bytes
```
nemesisdesign, line 509: during testing I realized that there are many different fields beginning with `rx_` and `tx_` (eg: `rx_packets`) that must be inverted, otherwise the data does not make sense.

I think the best way to solve this problem is to create another helper function to call here.

The helper function should:

- loop over all keys of `interface.statistics`
- skip if key does not start with `tx_` 
- from the current `tx_*` key, get the corresponding `rx_*` value and swap the two

Let me know if it's clear.

nemesisdesign (changes requested):

netjson-monitoring.lua
nemesisdesign: PS: I forgot to mention: let's also add a verbose comment that explains why we're doing this, for future reference

nemesisdesign (changes requested):
Nice, it works! But there's a minor style issue with the indentation, see below.

nemesisdesign (approved):
The indentation had tabs and spaces mixed, I fixed it.

Thanks! :+1:

nemesisdesign:
During testing I found out my assumption was wrong.
I think I understood now: is not that rx/tx are inverted, is that the interface tx/rx statistics of a wifi access point interface installed on a router have a different perspective: for the wifi interface, every packet that the user downloads is **sent** to the user, while every packet that the user uploads is received by the interface, that's why it is inverted. I do not think it has to do anything with the fact that the interface is bridged: a tap (layer2 VPN) member of a bridge which is used to tunnel traffic still shows received packets greater than sent packets, so that's why I changed the code further in https://github.com/openwisp/lua-monitoring/commit/266d98a4ec6eebdede227a75509eb4d11d4735cc to ensure the inversion is done only on wifi acess points and I made clear in the comment that we do this to show data from the perspective of the user rather than the perspective of the user.

PR #30, opened 2021-05-23
[qa] Split netjson-monitoring and Format files using luacheck #28

### LINKED ISSUE #28: [qa] Add quality assurance checks

devkapilbansal:
Closes #28


openwrt-openwisp-monitoring/files/dhcp.lua
```diff
@@ -0,0 +1,43 @@
+-- retrieve dhcp leases
+local uci = require('uci')
+local uci_cursor = uci.cursor()
+local utils = require('utils')
+
+local dhcp_functions = {}
```
nemesisdesign, line 6: let's get rid of the `_functions` suffix here and everywhere else

nemesisdesign (changes requested):

openwrt-openwisp-monitoring/files/netjson-monitoring.lua
```diff
@@ -0,0 +1,215 @@
...
+        disk = resources.parse_disk_usage()
+    }
+}
+
+local dhcp_leases = dhcp.get_dhcp_leases()
+if not utils.is_table_empty(dhcp_leases) then
+    netjson.dhcp_leases = dhcp_leases
+end
+
+local neighbors = neighbors_functions.get_neighbors()
```
devkapilbansal, line 57: Used `_functions` here because there is already a variable with same name

openwrt-openwisp-monitoring/files/netjson-monitoring.lua
```diff
@@ -0,0 +1,215 @@
...
+local cjson = require('cjson')
+
+local ubus_lib = require('ubus')
+local ubus = ubus_lib.connect()
+if not ubus then
+    error('Failed to connect to ubusd')
+end
+
+local dhcp = require('dhcp')
+local interface_functions = require('interfaces')
```
devkapilbansal, line 13: both variables, `interface` and `interface_functions` are defined :confused:

nickberry17: How about `host_interface` for the variable identifier?

openwrt-openwisp-monitoring/files/netjson-monitoring.lua
nickberry17: Suggested variable identifier `local host_neighbors = neighbors_functions.get_neighbors()`

nemesisdesign (commented):
what about creating a data structure that will hold the functions of each lua submodule?
We do something similar although simpler in openwisp-config: https://github.com/openwisp/openwisp-config/blob/master/openwisp-config/files/lib/openwisp/utils.lua

We could add a file which does:

```lua
local monitoring = {}

monitoring.dhcp = require('dhcp')
monitoring.interfaces = require('interfaces')
monitoring.neighbors = require('neighbors')
monitoring.resources = require('resources')
monitoring.utils = require('utils')
monitoring.wifi = require('wifi')

return monitoring
```

Then we can just do:

```lua
monitoring = require('monitoring')

neighbors = monitoring.neighbors.get_neighbors()
```

nickberry17:
> what about creating a data structure that will hold the functions of each lua submodule?
> We do something similar although simpler in openwisp-config: https://github.com/openwisp/openwisp-config/blob/master/openwisp-config/files/lib/openwisp/utils.lua
> 
> We could add a file which does:
> 
> ```lua
> local monitoring = {}
> 
> monitoring.dhcp = require('dhcp')
> monitoring.interfaces = require('interfaces')
> monitoring.neighbors = require('neighbors')
> monitoring.resources = require('resources')
> monitoring.utils = require('utils')
> monitoring.wifi = require('wifi')
> 
> return monitoring
> ```
> 
> Then we can just do:
> 
> ```lua
> monitoring = require('monitoring')
> 
> neighbors = monitoring.neighbors.get_neighbors()
> ```

This is the way it's done in LuCI, by the way.

openwrt-openwisp-monitoring/files/lib/openwisp/interfaces.lua
```diff
@@ -42,7 +42,7 @@ local specialized_interfaces = {
         local signal = io.popen('mmcli --output-json -m '..modem..' --signal-get'):read()
         if signal and pcall(function () signal = cjson.decode(signal) end) then
             -- only send data if not empty to avoid generating too much traffic
-            if not utils.is_table_empty(signal.modem) and not utils.is_table_empty(signal.modem.signal) then
+            if not monitoring.utils.is_table_empty(signal.modem.signal) then
```
devkapilbansal: Changed this as if `signal.modem` is empty then `signal.modem.signal` will point to nil

nemesisdesign: why?

devkapilbansal: I remembered it wrong. It will raise error. Reverting

openwrt-openwisp-monitoring/files/lib/openwisp/neighbors.lua
```diff
@@ -0,0 +1,70 @@
...
+                mac = mac,
+                interface = dev,
+                state = state
+            })
+        end
+    end
+    return arp_info
+end
+
+
```
nemesisdesign, line 58: why 2 blank lines?

nemesisdesign (changes requested):
Looks good, but I haven't tested it yet.

Please add a `./run-qa-checks` script for consistency with the other openwisp modules (please read the final sentence of [How to commit your changes properly](https://openwisp.io/docs/developer/contributing.html#how-to-commit-your-changes-properly) to understand why this is needed) and ensure this script is called from the CI and that is also documented in the README.

devkapilbansal:
> Looks good, but I haven't tested it yet.
> 
> Please add a `./run-qa-checks` script for consistency with the other openwisp modules (please read the final sentence of [How to commit your changes properly](https://openwisp.io/docs/developer/contributing.html#how-to-commit-your-changes-properly) to understand why this is needed) and ensure this script is called from the CI and that is also documented in the README.

We then need to introduce this in `openwisp-config` too. :thinking:

install-dev.sh
```diff
@@ -0,0 +1,8 @@
+set -e
```
feckert, line 1: Please add a valid file header and a shell shebang. `#!/bin/sh`

openwrt-openwisp-monitoring/files/lib/openwisp/interfaces.lua
```diff
@@ -0,0 +1,199 @@
...
+local interface_data = ubus:call('network.interface', 'dump', {})
+
+local interfaces = {}
+
+local specialized_interfaces = {
+    modemmanager = function(_, interface)
+        local modem = uci_cursor.get('network', interface['interface'], 'device')
+        local info = {}
+
+        local general = io.popen('mmcli --output-json -m '..modem):read("*a")
```
feckert, line 25: @nemesisdesign Why do we have to force usind modem manger here? Wouldn't it be better if we were more flexible here? On our system we use the openwrt proto 3g, uqmi or mbim and not the modem-manger.

feckert (commented):
We should provide all files with a valid file header where the license and the creator of the file can be seen.

openwrt-openwisp-monitoring/files/lib/openwisp/interfaces.lua
nemesisdesign: @feckert it would be great to support other ways other than modem.manager to get this information!
I suggest to discuss this in a separate issue since it's not something that Kapil can do because it is out of scope for this project.
Could you open a new issue and give some info on how to retrieve signal data with the packages you're currently using and what kind of output is returned?

If this is something your company would be interesting in using, we could invest some time in it.
Right now we are able to visualize this information in the device status tab, and in https://github.com/openwisp/openwisp-monitoring/pull/294 we're working on generating a few charts as well.

feckert: @nemesisdesign done https://github.com/openwisp/openwrt-openwisp-monitoring/issues/36

install-dev.sh
devkapilbansal: Hi @feckert I am not sure if we need to add file headers in all files. License file is added in each OpenWISP module but I am not able to find a similar file header in other modules.

openwrt-openwisp-monitoring/files/sbin/netjson-monitoring.lua
```diff
@@ -1,251 +1,34 @@
...
+local ubus = ubus_lib.connect()
 if not ubus then
     error('Failed to connect to ubusd')
 end
 
+local monitoring = require('openwisp.monitoring')
+
 -- helpers
-iwinfo_modes = {
+local iwinfo_modes = {
```
nemesisdesign, line 15: why don't we move this to `wifi.lua`?

openwrt-openwisp-monitoring/files/sbin/netjson-monitoring.lua
```diff
@@ -501,24 +105,24 @@ for radio_name, radio in pairs(wireless_status) do
...
-                netjson_interface.wireless.clients = netjson_clients(clients, is_mesh)
+            if not monitoring.utils.is_table_empty(clients) then
+                netjson_interface.wireless.clients = monitoring.wifi.netjson_clients(clients, is_mesh)
             end
             wireless_interfaces[name] = netjson_interface
         end
     end
 end
 
-function needs_inversion(interface)
```
nemesisdesign, line 517: this could also be part of `wifi.lua`, same for `invert_rx_tx`

nemesisdesign (changes requested):
Good progress @devkapilbansal!

I have generated NetJSON output with the following command bot with the previous scripts and the new ones in this PR:

`netjson-monitoring *`

The only problem I found is that with the new code, statistics of all interfaces are not included in the output, I think most likely you forgot to port this change:

- https://github.com/openwisp/openwrt-openwisp-monitoring/commit/2ae2def791fbbeaead26116e5a3a805310098517

So please port that one, as well as the commit coming after that, see the rest of my comments below.

nemesisdesign (approved):

Issue #31, opened 2021-05-23
[enhancement] Send offline data collected to server


devkapilbansal:
Depends on #29 https://github.com/openwisp/openwisp-monitoring/issues/298

- [x] Check if temporary data is available to send
- [x] Send data to server with explicit datetime
- [x] Delete data once request is successfule

Update:-

- [x] Specify in URL if data sent is old or new (no need to update device status at each point)

devkapilbansal:
Closed by b55b3e2daaefec68c12c58edc17452759f80df16

PR #25, opened 2021-04-01
[openwrt] Convert into an OpenWRT package #6

devkapilbansal:
Closes #6

PR #34, opened 2021-05-27
[enhancement] Temporarily store and send data

### LINKED ISSUE #29: [openwrt] Store data temporarily if HTTP request is unsuccessful
If daemon fails to send data to the server, store data in temporary directory with its date and time

Update:-

- [x] Break the procd service in two services, one to collect data and other to send data
- [x] Have a retry mechanism in sending data (exponential backoff) [link 1](https://coderwall.com/p/--eiqg/exponential-backoff-in-bash) [link 2](https://gist.github.com/nathforge/62456d9b18e276954f58eb61bf234c17)
- [x] #45

devkapilbansal:
Closes #29


devkapilbansal:
Blocked by #25

README.rst
```diff
@@ -31,6 +31,7 @@ UCI configuration options must go in ``/etc/config/monitoring``.
 
 - ``monitored_interfaces``: interfaces that needs to be monitored, defaults to ``*`` for all interfaces.
 - ``interval``: time after which device data should be sent to server, defaults to ``300``
+- ``required_memory``: free memory required to save data temporarily, defaults to ``0.05``
```
nemesisdesign, line 34: is it free or available memory?

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -31,6 +32,25 @@ fi
 
 URL="$BASE_URL/api/v1/monitoring/device/$UUID/?key=$KEY"
 
+tmp_dir="/tmp/monitoring"
```
nemesisdesign, line 35: what about `/tmp/openwisp/monitoring`?

nemesisdesign (commented):
I see you're saving the data, but when are you going to send it?

README.rst
devkapilbansal: Free memory

openwrt-openwisp-monitoring/files/monitoring.agent
devkapilbansal: I will change that

devkapilbansal:
> I see you're saving the data, but when are you going to send it?

I will send the data once post request is successful #31

README.rst
nemesisdesign: there's a problem with this: on modern systems it will always look like the free memory is almost nonexistent.
Linux allocates almost the entire memory to the operating system, ready to be used if needed.
That's why on modern systems you will see that there's also another metric called "available memory".
We should try to use available memory and use free memory only if missing.
Otherwise, we risk this work to be uneffective, because the daemon will think the memory is full and will not store the data.

nemesisdesign (changes requested):
There's also a conflict to resolve.

README.rst
devkapilbansal: My bad it is available memory. I mentioned free here as that memory used can be freed when needed :thinking:

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -31,6 +32,25 @@ fi
...
+
+check_available_memory(){
+	total=$(free | grep Mem | awk '{print $2}')
+	available=$(free | grep Mem | awk '{print $7}')
```
devkapilbansal, line 39: I take available memory here

feckert: You could use `ubus call system info` to get the information in one call

```
root@st-dev-07 ~ # ubus call system info
{
        "localtime": 1626076905,
        "uptime": 1443394,
        "load": [
                1760,
                1728,
                64
        ],
        "memory": {
                "total": 4117020672,
                "free": 3917463552,
                "shared": 2035712,
                "buffered": 33484800,
                "available": 3958448128,
                "cached": 62803968
        },
        "swap": {
                "total": 0,
                "free": 0
        }
}
```

devkapilbansal: Done

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -31,6 +32,25 @@ fi
 
 URL="$BASE_URL/api/v1/monitoring/device/$UUID/?key=$KEY"
 
+tmp_dir="/tmp/openwisp/monitoring"
```
nemesisdesign, line 35: use uppercase var name please

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -9,6 +9,7 @@ while [ -n "$1" ]; do
 		--verify_ssl) export verify_ssl="$2"; shift;;
 		--interval) export INTERVAL="$2"; shift;;
 		--monitored_interfaces) export MONITORED_INTERFACES=$2; shift;;
+		--required_memory) export required_percent=$2; shift;;
```
nemesisdesign, line 12: use upepr case var name please

nemesisdesign (changes requested):
Good progress, see my comments below.

I still don't see the code which sends the data to OpenWISP Monitoring, I know the API endpoints to receive the data there not ready yet, but you should start writing the logic which periodically sends it (you can then leave a stub for the POST request which sends it).

BTW: please keep in mind that the sending and the deletion of old data should happen only if the agent can communicate with the server, so probably it should happen if and after the current data is sent successfully, then for each point sent successfully the related data is deleted, iterating on each data point until all are sent and deleted.

devkapilbansal:
> Good progress, see my comments below.
> 
> I still don't see the code which sends the data to OpenWISP Monitoring, I know the API endpoints to receive the data there not ready yet, but you should start writing the logic which periodically sends it (you can then leave a stub for the POST request which sends it).
> 
> BTW: please keep in mind that the sending and the deletion of old data should happen only if the agent can communicate with the server, so probably it should happen if and after the current data is sent successfully, then for each point sent successfully the related data is deleted, iterating on each data point until all are sent and deleted.

Taking care of this in #31

nemesisdesign (changes requested):
> > Good progress, see my comments below.
> > I still don't see the code which sends the data to OpenWISP Monitoring, I know the API endpoints to receive the data there not ready yet, but you should start writing the logic which periodically sends it (you can then leave a stub for the POST request which sends it).
> > BTW: please keep in mind that the sending and the deletion of old data should happen only if the agent can communicate with the server, so probably it should happen if and after the current data is sent successfully, then for each point sent successfully the related data is deleted, iterating on each data point until all are sent and deleted.
> 
> Taking care of this in #31

Since we need both funcionalities (collecting data and sending) to be implemented in order to validate if this code really works, there's no reason to delay the implementation of the sending and work on it in a separate pull request, so please start working on it in this PR.
We can keep commits separate once we squash them, but definitely working on this in two PRs is only going to add overhead.

devkapilbansal:
> Since we need both funcionalities (collecting data and sending) to be implemented in order to validate if this code really works, there's no reason to delay the implementation of the sending and work on it in a separate pull request, so please start working on it in this PR.
> We can keep commits separate once we squash them, but definitely working on this in two PRs is only going to add overhead.

Sure, I will work on the sending part in this PR only

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -78,24 +65,126 @@ collect_data(){
...
+        mkdir -p "$TMP_DIR"
+        #save data with file_namecd
+        echo "$data" > "$TMP_DIR/$file_name"
+        [ "$verbose_mode" = "1" ] && logger -s "Data saved temporarily" \
+                                            -p daemon.info
+    fi
+    # get process id of container sending data
+    pid=$(ps | grep openwisp_monitoring | grep uuid | awk '{print $1}')
+    kill -SIGUSR1 "$pid"
+    sleep "$INTERVAL"
```
devkapilbansal, line 117: Sending SIGUSR1 signal once data is collected

devkapilbansal (commented):
If I am not wrong, this is what we need :point_down:

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -78,24 +65,126 @@ collect_data(){
...
+      done
+    done
+  done
+}
+
+if [ "$COLLECT" -eq "1" ]; then
+  MONITORED_INTERFACES=${MONITORED_INTERFACES:-*}
+  save_data
+else
+  trap send_data USR1
```
devkapilbansal, line 185: Trapping send_data to ensure data is sent as soon as it is collected

devkapilbansal (commented):
If I am not wrong, this is what we need :point_down:

README.rst
```diff
@@ -40,6 +40,8 @@ UCI configuration options must go in ``/etc/config/monitoring``.
 - ``monitored_interfaces``: interfaces that needs to be monitored, defaults to ``*`` for all interfaces.
 - ``interval``: time after which device data should be sent to server, defaults to ``300``.
 - ``verbose_mode``: can be enabled (set to ``1``) to ease `debugging <#debugging>`__ in case of issues, defaults to ``0`` (disabled).
+- ``required_memory``: available memory required to save data temporarily, defaults to ``0.05``.
+- ``max_retries``: number of time the service should try to send data to server in case of failure, defaults to ``5``.
```
nemesisdesign, line 45: I think we need to add some more info in the README to explain how the 2 processes work.

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -78,24 +65,139 @@ collect_data(){
...
+save_data() {
+  # Remove double quotes from interfaces
+  MONITORED_INTERFACES=$(echo "$MONITORED_INTERFACES" | tr -d '"')
+
+  while true
+  do
+    data="$(collect_data)"
+    memory_available="$(check_available_memory)"
+    if [ "$memory_available" != "0" ]; then
+        file_name="$(date +'%d-%m-%Y_%H:%M:%S').000000.json" # extra zeroes are added for nanoseconds precision
```
devkapilbansal, line 106: Made this change for openwisp-monitoring module.

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -78,24 +65,140 @@ collect_data(){
...
+save_data() {
+  # Remove double quotes from interfaces
+  MONITORED_INTERFACES=$(echo "$MONITORED_INTERFACES" | tr -d '"')
+
+  while true
+  do
+    data="$(collect_data)"
+    memory_available="$(check_available_memory)"
+    if [ "$memory_available" != "0" ]; then
+        file_name="$(date -u +'%d-%m-%Y_%H:%M:%S').000000.json" # extra zeroes are added for nanoseconds precision
```
nemesisdesign, line 107: please move the comment above the line of code to avoid the line being too long

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -78,24 +65,140 @@ collect_data(){
...
+        # make directory
+        mkdir -p "$TMP_DIR"
+        #save data with file_name
```
nemesisdesign, line 110: always leave a space after the comment

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -78,24 +65,140 @@ collect_data(){
...
+  URL="$BASE_URL/api/v1/monitoring/device/$UUID/?key=$KEY"
+
+  CURL_COMMAND="curl -s -w "%{http_code}""
+  [ "$VERIFY_SSL" = "0" ] && CURL_COMMAND="$CURL_COMMAND -k"
+  [ "$verbose_mode" = "1" ] && CURL_COMMAND="$CURL_COMMAND -v"
+  echo "0"
+}
+
+save_data() {
+  # Remove double quotes from interfaces
```
nemesisdesign, line 99: if you're starting comments with lowercase letters be consistent and ensure the style of all comments is consistent

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -26,40 +30,23 @@ if [ "$VERSION" -eq "1" ]; then
...
-         -t openwisp_monitoring \
-         -p daemon.err
-  exit 1
-fi
-
-# Remove double quotes from interfaces
-MONITORED_INTERFACES=$(echo "$MONITORED_INTERFACES" | tr -d '"')
+check_available_memory(){
+  total=$(ubus call system info | grep -m1 total | tr -cd 0-9)
+  available=$(ubus call system info | grep available | tr -cd 0-9)
```
nemesisdesign, line 39: can you use jsonfilter to get this? It should be available by default on OpenWRT and should result in being cleaner.

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -1,6 +1,7 @@
 #!/bin/sh
 
 VERSION=0 #default value of version
+COLLECT=0
```
nemesisdesign, line 4: wouldn't it be more readable if we had something like `MODE="collect"` and `MODE="send"`?

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -78,24 +65,140 @@ collect_data(){
...
+  echo "0"
+}
+
+save_data() {
+  # Remove double quotes from interfaces
+  MONITORED_INTERFACES=$(echo "$MONITORED_INTERFACES" | tr -d '"')
+
+  while true
+  do
+    data="$(collect_data)"
```
nemesisdesign, line 104: why are you collecting the data before checking for the memory availability?
If the data is large and the memory is already at full capacity, isn't there's a chance the agent my crash here? In that case it would not be as robust as intended wouldn't it?

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -78,24 +65,140 @@ collect_data(){
...
+        sleep "$INTERVAL" & wait $!
+        trap "" USR1
+        continue
+      fi
+      basefilename=${file##*/}
+      filename=${basefilename%.*}
+      url="$URL&time=$filename"
+      data=$(cat "$file")
+      # retry sending data in case of failure
+      MAX_RETRY=${MAX_RETRY:-5}
```
nemesisdesign, line 152: 2 observations here:

- isn't this value a constant which doesn't change in the loop?
  if so, why define it inside loop, which is executed many times? Can't this be defined just once at the beginning of the function?
- I think it can be renamed to `MAX_RETRIES`

Keep in mind the same suggestion for any other variable which doesn't need to be in the loop, I see a few different ones here so please double check here and in any other loop.

Please also keep in mind that uppercase is generally used for constants which are defined at the beginning of the program and used in different places. For anything else defined only in specific functions, please use lowercase naming and flag them as [local variables](https://tldp.org/LDP/abs/html/localvar.html).

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -78,24 +65,140 @@ collect_data(){
...
+      FAILURES=0
+      BASE_SECONDS=1
+      # check if the data is latest or old one
+      [ "$(echo "$TMP_DIR"/*.json | awk '{print $2}')" ] || url="$url&current=true"
+      while true
+      do
+        FAILURES=$((FAILURES+1))
```
nemesisdesign, line 159: why are you adding 1 here, before any failure?
Is this `FAILURES` variable actually a counter of how many attempts have been made? In that case I would name it `attempts`.

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -78,24 +65,140 @@ collect_data(){
...
+          break
+        fi
+        #send data
+        response_code=$($CURL_COMMAND -H "Content-Type: application/json" \
+          -d "$data" \
+          -v "$url")
+        if [ "$response_code" = "200" ]; then
+          [ "$verbose_mode" = "1" ] && logger -s "Data sent successfully." \
+                                              -p daemon.info
+          #Removed saved data
```
nemesisdesign, line 182: please always leave a space from the beginning of the comment and keep the first case consistent with the other comments. Applies to any other comment.

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -78,24 +65,140 @@ collect_data(){
...
+          [ "$verbose_mode" = "0" ] && logger -s "Run with verbose mode to find more." \
+                                              -t openwisp_monitoring \
+                                              -p daemon.err
+          break
+        fi
+        #send data
+        response_code=$($CURL_COMMAND -H "Content-Type: application/json" \
+          -d "$data" \
+          -v "$url")
+        if [ "$response_code" = "200" ]; then
```
nemesisdesign, line 179: we not use -eq instead of = for consistency with other sections? Let's choose one way of doing it and let's stick to it, however I personally prefer to use `==` instead of `=` because the single equal sign can be confused with assignment.

Applies to all other boolean evaluations.

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -78,24 +65,140 @@ collect_data(){
...
+          [ "$verbose_mode" = "1" ] && logger -s "Data sent successfully." \
+                                              -p daemon.info
+          #Removed saved data
+          rm "$file"
+          break
+        else
+          TIME=$((BASE_SECONDS * 2 **(FAILURES-1)))
+          [ "$verbose_mode" = "1" ] && logger -s "Data not sent successfully. Retrying after $TIME seconds" \
```
nemesisdesign, line 187: I'd use "Retrying in $TIME seconds"

README.rst
```diff
@@ -40,6 +40,18 @@ UCI configuration options must go in ``/etc/config/monitoring``.
 - ``monitored_interfaces``: interfaces that needs to be monitored, defaults to ``*`` for all interfaces.
 - ``interval``: time after which device data should be sent to server, defaults to ``300``.
 - ``verbose_mode``: can be enabled (set to ``1``) to ease `debugging <#debugging>`__ in case of issues, defaults to ``0`` (disabled).
+- ``required_memory``: available memory required to save data temporarily, defaults to ``0.05``.
```
nemesisdesign, line 43: what is 0.05, a percentage? Please make it explicit.

README.rst
```diff
@@ -40,6 +40,18 @@ UCI configuration options must go in ``/etc/config/monitoring``.
...
+- ``max_retries``: number of time the service should try to send data to server in case of failure, defaults to ``5``.
```
nemesisdesign, line 44: suggestion:

maximum number of retries in case of failures to send data to the server, defaults to ``5``.
in case the maximum number of retries is reached .... (link the section below so the user knows what happens in this case).

README.rst
```diff
@@ -40,6 +40,18 @@ UCI configuration options must go in ``/etc/config/monitoring``.
...
+
+OpenWISP Monitoring Agent
+-------------------------
```
nemesisdesign, line 47: This title is not very informative of what is being explained here.

We want to explain that this agent works with 2 processes, one for collecting and one for sending.

We could name this section "Collecting vs sending".

README.rst
```diff
@@ -40,6 +40,18 @@ UCI configuration options must go in ``/etc/config/monitoring``.
...
+
+We use two procd services in `monitoring agent <https://github.com/openwisp/openwrt-openwisp-monitoring/blob/master/openwrt-openwisp-monitoring/files/monitoring.agent>`_, one for collecting the data and other for sending the data.
+
+**SIGUSR1** signals are used to instantly send the data when collected. However, the service will keep trying
+to send data periodically.
+
+**Note:-** It may be possible that the server is down for a long time, then sending data may take time then the interval after which data is collected creating a gap in data points. To avoid this, two services are used.
```
nemesisdesign, line 54: Citing the server as being down is not a great example, the most common case is the OpenWRT device simply not being able to send the data for network connectivity issues.

I think we could divide this section further and explain what happens in the most important cases, eg:

1st heading: collecting vs sending

Intro: this agent uses 2 different processes, one for sending data and one for collecting it.
This allows to handle failures in sending the data in a more flexible way because old data stored during network connectivity issues can be sent at the same time while new data is being collected: if old data has piled up for sending and takes several minutes to be uploaded, new data is collected without being delayed by the sending.

Subheading 1: Collect mode

In this mode the agent is in charge of collecting data ...
The data is collected only if ...
If there is not enough memory ...
Once data is collected, a signal is sent to the other agent so that ...

Subheading 2: Send mode

In this mode the agent .....
If sending fails .... explain retries...
If all attempts fail ...

README.rst
```diff
@@ -96,7 +108,6 @@ you will need to select the *openwisp-monitoring* variant and *netjson-monitorin
     ./scripts/feeds update -a
     ./scripts/feeds install -a
     make menuconfig
-    # go to Base system, then select rpcd
```
nemesisdesign, line 99: are you sure this is needed and are we sure this section of explainin how to select the packages with makemenu config is needed at all? The explanations above are enough in most cases, this seems redundant to me.

nemesisdesign (changes requested):
Good progress @devkapilbansal, please see my comments below.

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -78,24 +65,140 @@ collect_data(){
...
+    if [ "$memory_available" != "0" ]; then
+        file_name="$(date -u +'%d-%m-%Y_%H:%M:%S').000000.json" # extra zeroes are added for nanoseconds precision
+        # make directory
+        mkdir -p "$TMP_DIR"
+        #save data with file_name
+        echo "$data" > "$TMP_DIR/$file_name"
+        [ "$verbose_mode" = "1" ] && logger -s "Data saved temporarily" \
+                                            -p daemon.info
+    fi
+    # get process id of container sending data
```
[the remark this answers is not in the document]
devkapilbansal: :grimacing: Fixed now

openwrt-openwisp-monitoring/files/monitoring.agent
devkapilbansal: Changing in to increase after failure

openwrt-openwisp-monitoring/files/monitoring.agent
devkapilbansal: `==` is not defined in strict POSIX. That's why I used single `=` as suggested by `shellcheck`.

Changing it to -eq where integer comparisions are made for consistency.

README.rst
devkapilbansal: This was not, that's why I removed it :sweat_smile:

devkapilbansal: I added this section to make it consistent with `**openwisp-config**

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -26,76 +29,184 @@ if [ "$VERSION" -eq "1" ]; then
...
+      file_name="$(date -u +'%d-%m-%Y_%H:%M:%S').000000.json"
+      # make directory 
+      mkdir -p "$TMP_DIR"
+      # save data with file_name 
+      echo "$data" > "$TMP_DIR/$file_name"
+      [ "$verbose_mode" -eq "1" ] && logger -s "Data saved temporarily" \
+                                          -p daemon.info
+    fi
+    # get process id of the process sending data 
+    pid=$(pgrep -f "openwisp_monitoring.*uuid")
```
devkapilbansal, line 113: We also don't need to evaluate process id each time but in starting collect data will starts first before send data :thinking:

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -26,76 +29,184 @@ if [ "$VERSION" -eq "1" ]; then
...
-fi
-
-# Remove double quotes from interfaces
-MONITORED_INTERFACES=$(echo "$MONITORED_INTERFACES" | tr -d '"')
-
-URL="$BASE_URL/api/v1/monitoring/device/$UUID/?key=$KEY"
+check_available_memory(){
+  total=$(ubus call system info | jsonfilter -e '@.memory.total')
+  available=$(ubus call system info | jsonfilter -e '@.memory.available')
+  required=$(echo - |awk -v percent="$REQUIRED_PERCENT" -v total="$total" '{printf("%.f",percent*total)}')
```
devkapilbansal, line 39: Is there a way to improve this too? I tried searching for one but end in futile.

devkapilbansal (commented):
Made some minor improvements in the agent as suggested above and will add a commit for documentation soon.
I have some doubts here :point_down:

openwrt-openwisp-monitoring/files/monitoring.agent
nemesisdesign: but the other process may be restarted

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -167,38 +161,52 @@ send_data() {
...
   MONITORED_INTERFACES=${MONITORED_INTERFACES:-*}
+  # remove double quotes from interfaces 
+  MONITORED_INTERFACES=$(echo "$MONITORED_INTERFACES" | tr -d '"')
   save_data
-else
+elif [ "$MODE" = "send" ]; then
   VERIFY_SSL=${VERIFY_SSL:-0}
   set_url_and_curl && send_data
+else
+  logger -s "Please specify correct mode" \
```
nemesisdesign, line 210: instead of logging an error, I suggest just echoing to standard error a message like "The supplied mode is invalid, only collect and send are accepted."

README.rst
nemesisdesign: lol, I think it may be a good idea to remove it from there too then :grin: 

hey what do you think @nickberry17 @feckert, do you think this is useful at all?

openwrt-openwisp-monitoring/files/monitoring.agent
nemesisdesign: at the moment I don't have ideas on how to further improve this.

README.rst
```diff
@@ -40,6 +40,46 @@ UCI configuration options must go in ``/etc/config/monitoring``.
...
+- ``max_retries``: maximum number of retries in case of failures to send data to server in case of failure, defaults to ``5`` retries.
+
+In case, `maximum retries are reached <#send-mode>`_, agent will try sending data again in next cycle.
+
+Collecting vs Sending
+---------------------
+
+We use two procd services in `monitoring agent <https://github.com/openwisp/openwrt-openwisp-monitoring/blob/master/openwrt-openwisp-monitoring/files/monitoring.agent>`_, one for collecting the data and other for sending the data.
+
+This helps handle failure in sending the data in more flexible way. Old data saved during network connectivity issues can be sent while new data is being collected. If old data has piled up and takes several minutes to be uploaded, new data will be collected without waiting for the sending to complete.
```
nemesisdesign, line 53: shouldn't it be "helps to handle"?

README.rst
```diff
@@ -40,6 +40,46 @@ UCI configuration options must go in ``/etc/config/monitoring``.
...
+If there is no data file, the agent will sleep for the time interval and check for data file again. This will be continued until a data file is found.
+If a signal is received from other agent, then the sleep will be interrupted and agent will start sending data.
+
+If agent fails to send data to the server, exponential backoff will be used to retry until max retries are reached.
+If all attempts of sending data failed, the agent will try to send data in next cycle.
+
+If data is sent successfully, then the data file will be deleted and agent will look for another file.
+
+**SIGUSR1** signals are used to instantly send the data when collected. However, the service will keep trying
+to send data periodically.
```
nemesisdesign, line 82: @nickberry17 your review of this text would be super helpful!

nemesisdesign (commented):
Looks good, but I need to test it still (couldn't test it because it seems the device currently available in our test system doesn't have the agent, I left some hints in the chat channel).

@feckert @nickberry17 your input in this PR will be much appreciated :pray:

nemesisdesign (changes requested):
Good progress @devkapilbansal, I tested it quickly and it seems to me that the basic goals are reached, so :+1:  :clap: for that! 

Here's the issues I found during testing.
I haven't looked into OpenWISP Monitoring yet.

### Rule of silence is broken

**Rule of Silence: When a program has nothing surprising to say, it should say nothing.**
http://www.catb.org/~esr/writings/taoup/html/ch01s06.html#id2878450

Looking at the log, I see:

```
Fri Aug 13 10:30:24 2021 daemon.info openwisp_monitoring: SIGUSR1 received! Sending data
Fri Aug 13 10:35:25 2021 daemon.info openwisp_monitoring: SIGUSR1 received! Sending data
Fri Aug 13 10:40:26 2021 daemon.info openwisp_monitoring: SIGUSR1 received! Sending data
```

This would fill any log collector with useless data, let's avoid this unless in verbose mode.

The same is valid when there's something wrong: we do not need to log every failed request when not in verbose mode, we can simply log the first failure and let the user know the sending is not working, then when the sending works again we can log a line saying sending is working again and that's it.

### Name of files stored could be shortened to save memory

I see the filenames are something like:

`13-08-2021_01:39:29.000000.json`

But I think the last part is unnecessarily eating the memory of the device.

Can you add `.000000` programmatically instead of having to use it in the file name?

Then, I think we can live without `.json`, isn't it?

devkapilbansal:
> ```
> Fri Aug 13 10:30:24 2021 daemon.info openwisp_monitoring: SIGUSR1 received! Sending data
> Fri Aug 13 10:35:25 2021 daemon.info openwisp_monitoring: SIGUSR1 received! Sending data
> Fri Aug 13 10:40:26 2021 daemon.info openwisp_monitoring: SIGUSR1 received! Sending data
> ```
> 
> This would fill any log collector with useless data, let's avoid this unless in verbose mode.

:grimacing: Fixed now

> 
> The same is valid when there's something wrong: we do not need to log every failed request when not in verbose mode, we can simply log the first failure and let the user know the sending is not working, then when the sending works again we can log a line saying sending is working again and that's it.

:thinking: Would it be like maintaining a variable to know when the request start failing and when the process is normal again. Like, logging only when the value of variable change

> ### Name of files stored could be shortened to save memory
> I see the filenames are something like:
> 
> `13-08-2021_01:39:29.000000.json`
> 
> But I think the last part is unnecessarily eating the memory of the device.
> 
> Can you add `.000000` programmatically instead of having to use it in the file name?

I added it in URL directly. :+1: 

> 
> Then, I think we can live without `.json`, isn't it?

I added `.json` previously to distinguish data files from any other file that we may save in future. Removed now.

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -26,76 +29,179 @@ if [ "$VERSION" -eq "1" ]; then
...
-
-CURL_COMMAND="curl -s -w "%{http_code}""
-[ "$VERIFY_SSL" = "0" ] && CURL_COMMAND="$CURL_COMMAND -k"
-[ "$verbose_mode" = "1" ] && CURL_COMMAND="$CURL_COMMAND -v"
+  if [ "$available" -ge "$required" ]; then
+    echo "1"
+  else
+    [ "$verbose_mode" -eq "1" ] && logger -s "Not enough memory available" \
+                                          -p daemon.err
+    echo "0"
```
devkapilbansal, line 48: I think this should be changed too.
0 should be returned if case of success, else we should return 1

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -26,76 +29,179 @@ if [ "$VERSION" -eq "1" ]; then
...
+        trap handle_sigusr1 USR1
+        # SIGUSR1 signal received, interrupt sleep and continue sending data
+        sleep "$INTERVAL" & wait $!
+        trap "" USR1
+        continue
+      fi
+      basefilename=${file##*/}
+      filename=${basefilename%.*}
+      # extra zeroes are added for nanoseconds precision
+      url="$URL&time=$filename.000000"
```
devkapilbansal, line 139: Added zeroes while sending data in url directly.

openwrt-openwisp-monitoring/files/monitoring.agent
devkapilbansal: Changed

openwrt-openwisp-monitoring/files/monitoring.agent
nemesisdesign: can we use -eq here too?

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -26,76 +29,180 @@ if [ "$VERSION" -eq "1" ]; then
...
+    done
+  done
+}
+
+if [ -z "$MODE" ]; then
+  echoerr "missing required --mode option"
+  exit 1
+fi
+
+if [ "$MODE" = "collect" ]; then
```
nemesisdesign, line 197: can we use -eq here too please?

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -26,76 +29,180 @@ if [ "$VERSION" -eq "1" ]; then
...
+  MONITORED_INTERFACES=${MONITORED_INTERFACES:-*}
+  # remove double quotes from interfaces 
+  MONITORED_INTERFACES=$(echo "$MONITORED_INTERFACES" | tr -d '"')
+  save_data
+elif [ "$MODE" = "send" ]; then
```
nemesisdesign, line 202: can we use -eq here too please?

nemesisdesign (changes requested):
It's working a lot better, see my comments below, we should be near to merge this.

nemesisdesign (changes requested):
I just noticed two issues.

1. error message is shown more than once in non verbose mode:

Thu Aug 19 02:03:08 2021 daemon.err openwisp_monitoring: Data not sent successfully. Response code is 000 Error message is  Run with verbose mode to find more.
Thu Aug 19 02:08:08 2021 daemon.err openwisp_monitoring: Data not sent successfully. Response code is 000 Error message is  Run with verbose mode to find more.
Thu Aug 19 02:13:09 2021 daemon.err openwisp_monitoring: Data not sent successfully. Response code is 000 Error message is  Run with verbose mode to find more.

2. the error message is ... 

I think this:

`Response code is 000 Error message is  Run with verbose mode to find more.`

Would be better written like this:

`Response code is "000". Error message is "". Run with verbose mode to find more.`

devkapilbansal:
> I just noticed two issues.
> 
> 1. error message is shown more than once in non verbose mode:
> 
> ```
> Thu Aug 19 02:03:08 2021 daemon.err openwisp_monitoring: Data not sent successfully. Response code is 000 Error message is  Run with verbose mode to find more.
> Thu Aug 19 02:08:08 2021 daemon.err openwisp_monitoring: Data not sent successfully. Response code is 000 Error message is  Run with verbose mode to find more.
> Thu Aug 19 02:13:09 2021 daemon.err openwisp_monitoring: Data not sent successfully. Response code is 000 Error message is  Run with verbose mode to find more.
> ```
> 
> 1. the error message is ...
> 
> I think this:
> 
> `Response code is 000 Error message is Run with verbose mode to find more.`
> 
> Would be better written like this:
> 
> `Response code is "000". Error message is "". Run with verbose mode to find more.`

This error message is coming from #63 that I included with this to test. Will fix it there

nemesisdesign:
> > I just noticed two issues.
> > 
> > 1. error message is shown more than once in non verbose mode:
> > 
> > ```
> > Thu Aug 19 02:03:08 2021 daemon.err openwisp_monitoring: Data not sent successfully. Response code is 000 Error message is  Run with verbose mode to find more.
> > Thu Aug 19 02:08:08 2021 daemon.err openwisp_monitoring: Data not sent successfully. Response code is 000 Error message is  Run with verbose mode to find more.
> > Thu Aug 19 02:13:09 2021 daemon.err openwisp_monitoring: Data not sent successfully. Response code is 000 Error message is  Run with verbose mode to find more.

I was thinking, since in openwisp-config I noticed there's a log line every time there's a failure, I think this is fine.
It may be more useful you focus on other things.

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -10,7 +10,10 @@ while [ -n "$1" ]; do
     --verify_ssl) export VERIFY_SSL="$2"; shift;;
     --interval) export INTERVAL="$2"; shift;;
     --monitored_interfaces) export MONITORED_INTERFACES="$2"; shift;;
-    --verbose_mode) export verbose_mode=$2; shift;;
+    --verbose_mode) export verbose_mode="$2"; shift;;
```
nemesisdesign, line 13: verbose_mode is still lowercase

nemesisdesign (changes requested):

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -26,76 +29,178 @@ if [ "$VERSION" -eq "1" ]; then
...
+            logger -s "Data sent successfully" \
+                   -t openwisp_monitoring \
+                   -p daemon.info
+            FAILING=0
+          fi
+          # remove saved data
+          rm "$file"
+          break
+        else
+          TIME=$((BASE_SECONDS * 2 **(failures)))
```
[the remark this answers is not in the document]
devkapilbansal: I was trying to get exponent instead of square, i.e. 2^failures. Nvm, I use a cleaner way to do so.

openwrt-openwisp-monitoring/files/monitoring.agent
devkapilbansal: Here, I use `=` beause in case of failure, error message will be there along with `response_code` which will be string. This is changed to `-eq` in #63 where error message is instead stored in a file.

nemesisdesign (approved):
Thank you very much @nickberry17 @devkapilbansal :+1:

README.rst
devkapilbansal: Would it be better to replace the section with a GIF?

OpenWISP Package Compiling

nemesisdesign: Yes, is always good to have something visual :+1: 
pelase upload the gif to a media branch (to keep the main branch lightweight).

nickberry17: > Would it be better to replace the section with a GIF?
> 
> OpenWISP Package Compiling

I like the idea of a GIF, but it add a note saying it's not necessary to select all of those packages; only the one you need.

PR #42, opened 2021-07-13
[fix] Fix errors in package and improve debugging

### LINKED ISSUE #41: [bug] Data not sent successfully. Exit code is 127
While testing the daeamon I get the following:

```
Mon Jul  5 00:23:06 2021 daemon.info monitoring: Collecting NetJSoN Monitoring data
Mon Jul  5 00:23:06 2021 daemon.err monitoring: Data not sent successfully. Exit code is 127
``` 

First of all, can you please change any occurrence of output or docs mentioning "NetJSON" which is not spelled this way and end ensure follows this spelling?

Then, `Data not sent successfully. Exit code is 127` is not super helpful in finding out what's wrong.

What steps can we take in order to improve the life of users which will be using this package?
Can we include the code line and command which are failing?

devkapilbansal:
Fixes #41

### DIFF
--- a/openwrt-openwisp-monitoring/files/monitoring.config
+++ b/openwrt-openwisp-monitoring/files/monitoring.config
@@ -1,3 +1,4 @@
 config monitoring 'monitoring'
 	option monitored_interfaces '"*"'
 	option interval '300'
+	option verbose_mode '0'
--- a/openwrt-openwisp-monitoring/files/monitoring.config
+++ b/openwrt-openwisp-monitoring/files/monitoring.config
@@ -1,3 +1,3 @@
 config monitoring 'monitoring'
-	option monitored_interfaces '*'
+	option monitored_interfaces '"*"'
 	option interval '300'
--- a/openwrt-openwisp-monitoring/files/monitoring.config
+++ b/openwrt-openwisp-monitoring/files/monitoring.config
@@ -1,4 +1,4 @@
 config monitoring 'monitoring'
-	option monitored_interfaces '"*"'
+	option monitored_interfaces '*'
 	option interval '300'
 	option verbose_mode '0'
--- a/openwrt-openwisp-monitoring/files/monitoring.init
+++ b/openwrt-openwisp-monitoring/files/monitoring.init
@@ -40,6 +40,7 @@ start_service() {
   config_load openwisp_monitoring
   config_get monitored_interfaces monitoring monitored_interfaces "'*'"
   config_get interval monitoring interval "300"
+  config_get_bool verbose_mode monitoring verbose_mode "1"
 
   interval="$(time_to_seconds "$interval")"
   if [ "$interval" -lt 1 ]; then
@@ -50,10 +51,12 @@ start_service() {
   fi
   interval="--interval $interval"
   monitored_interfaces="--monitored_interfaces $monitored_interfaces"
+  verbose="--verbose_mode $verbose_mode"
 
   procd_open_instance "openwisp_monitoring_monitoring"
-  procd_set_param command $PROG $base_url $uuid $key $verify_ssl $interval $monitored_interfaces
+  procd_set_param command $PROG $base_url $uuid $key $verify_ssl $interval $monitored_interfaces $verbose
   procd_set_param respawn "${respawn_threshold:-3600}" "${respawn_timeout:-5}" "${respawn_retry:-5}"
+  procd_set_param stdout 1
   procd_set_param stderr 1
   procd_close_instance
   logger -s "$PROG_NAME started" \
--- a/openwrt-openwisp-monitoring/files/monitoring.init
+++ b/openwrt-openwisp-monitoring/files/monitoring.init
@@ -45,8 +45,8 @@ start_service() {
   interval="$(time_to_seconds "$interval")"
   if [ "$interval" -lt 1 ]; then
     logger -s "Interval is invalid. Use time value(eg: '10', '2m', '3h', '1d')" \
-         -t monitoring \
-         -p daemon.err
+           -t openwisp_monitoring \
+           -p daemon.err
     exit 1
   fi
   interval="--interval $interval"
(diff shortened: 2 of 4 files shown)


openwrt-openwisp-monitoring/files/monitoring.init
```diff
@@ -52,8 +52,9 @@ start_service() {
     monitored_interfaces="--monitored_interfaces $monitored_interfaces"
 
     procd_open_instance "openwisp_monitoring_monitoring"
-    procd_set_param command $PROG "$base_url" "$uuid" "$key" "$verify_ssl" "$interval" "$monitored_interfaces"
+    procd_set_param command $PROG $base_url $uuid $key $verify_ssl $interval $monitored_interfaces
     procd_set_param respawn "${respawn_threshold:-3600}" "${respawn_timeout:-5}" "${respawn_retry:-5}"
+    procd_set_param stderr 1
```
devkapilbansal, line 57: Adding this to log all errors in the instance so that it can be accessed using logread

openwrt-openwisp-monitoring/files/monitoring.init
```diff
@@ -52,8 +52,9 @@ start_service() {
     monitored_interfaces="--monitored_interfaces $monitored_interfaces"
 
     procd_open_instance "openwisp_monitoring_monitoring"
-    procd_set_param command $PROG "$base_url" "$uuid" "$key" "$verify_ssl" "$interval" "$monitored_interfaces"
+    procd_set_param command $PROG $base_url $uuid $key $verify_ssl $interval $monitored_interfaces
```
devkapilbansal, line 55: Reverting quotes as this was causing `monitoring.agent` to read `--url http://www.....` as one word

openwrt-openwisp-monitoring/files/monitoring.config
```diff
@@ -1,3 +1,4 @@
 config monitoring 'monitoring'
-	option monitored_interfaces '*'
+	option monitored_interfaces '"*"'
```
nemesisdesign, line 2: requiring the user to remember to wrap the value around double quotes is really bad and weird. I think we can do better than this.

openwrt-openwisp-monitoring/files/monitoring.init
```diff
@@ -50,10 +51,12 @@ start_service() {
     fi
     interval="--interval $interval"
     monitored_interfaces="--monitored_interfaces $monitored_interfaces"
```
nemesisdesign, line 53: does 

```
monitored_interfaces="--monitored_interfaces \"$monitored_interfaces\""
```

allow us to not having to wrap the value in quotes in the UCI config file?

nemesisdesign (changes requested):
With this patch, the daemon sends data to OpenWISP Monitoring :+1: .

Here's the issues I found during testing.

### Non verbose mode

I see these two lines in non-verbose mode:

```
Fri Jul 16 10:32:36 2021 daemon.info monitoring: Collecting NetJSON Monitoring data
Fri Jul 16 10:32:38 2021 daemon.info monitoring: Data sent successfully.
```

But this output should not be shown in non-verbose mode, otherwise logs will be filled with irrelevant lines.

### Traffic stats seem missing

I don't see traffic stats being collected on the device I am using for testing, once I installed this branch and remove the old monitoring scripts, the traffic charts are flat zero.

Did you test this?

### Verbose mode testing

Here's a few problems I see with verbose mode.

#### Identifeir different

The logread output uses different identifiers: monitoring, openwisp_monitoring, openwisp_monitoring[<number>]. This is bad because log collectors will interpret these as different programs, can you make sure this is always the same? `monitoring` seems to generic, I think it should be `openwisp_monitoring`.

#### Logged response can be improved

Here's an example output: `Data not sent successfully. Response code is {"detail":"Authentication credentials were not provided."}403.`. The message is readable but it's weird. The response body is `{"detail":"Authentication credentials were not provided."}` while the response status code is 403. Can you turn this into something like:
`The response received was: 403 {"detail":"Authentication credentials were not provided."}`?

#### Run with verbose mode redundant

This is showing up with verbose mode:

```
Fri Jul 16 10:53:31 2021 daemon.err monitoring: Run with verbose mode to find more.
Fri Jul 16 10:53:31 2021 daemon.err openwisp_monitoring[4023]: monitoring: Run with verbose mode to find more.
``` 

1. we're already in verbose mode so we don't need to send this output.
2. if we're not in verbose mode, ensure the output is not repeated twice
3. do you see how one log line is identified by `daemon.err monitoring` and the otehr one by `daemon.err openwisp_monitoring[4023]`? This was mentioned in one of the previous review sections but please make sure to avoid it.

devkapilbansal:
Thanks for testing this. I made some changes as follows:

**Non Verbose Mode**
Unnecessary lines have been removed. Those two lines will be logged in verbose mode only. 

**Traffic stats**
I just ensured that traffic stats are sent in netjson response from the scripts. Will test it to be more sure.

**Identifier**
All messages tagged with `monitoring` are changed to `openwisp_monitoring`. But in verbose_mode, messages are coming as `openwisp_monitoring[number]` and I am unsure of that number. I think that this is the id of procd instance. :thinking: 

**Redundant Log Messages**
Messages were logged twice in verbose mode. This was because messages were logged once by the logger tagged with the identifier we were passing.These messages were catched by the procd instance and logged again with procd instance name as the identifier.
For this, I removed the identifier from the messages I was logging in verbose mode so that we get relevant lines only when grepping log messages.
If we use `logread` then the messages are redundant in verbose_mode but passing `logread | grep openwisp_monitoring` will show the message once.

nemesisdesign (approved):
I did a couple of minor improvements https://github.com/openwisp/openwrt-openwisp-monitoring/pull/42/commits/021307c57289763847e3f3f59dda051ab1751d9e https://github.com/openwisp/openwrt-openwisp-monitoring/pull/42/commits/0083e00d3824833aea1de0062e84f2be97acf7e3.

Tested and it's working better, I think there's room for improvements but let's merge and I will open new issues for improvements.

PR #48, opened 2021-08-09
[fix] Modified functions using popen #15 #32

### LINKED ISSUE #15: [bug] get_neighbors() may fail with "Interrupted system call"
```
lua: /usr/sbin/netjson-monitoring:67: Interrupted system call
stack traceback:
	[C]: in function '(for generator)'
	/usr/sbin/netjson-monitoring:67: in function 'get_ip_neigh'
	/usr/sbin/netjson-monitoring:84: in function 'get_neighbors'
	/usr/sbin/netjson-monitoring:181: in main chunk
	[C]: ?
```

I need to test this on more devices before ruling out if it's a system issue or something that has to be dealt with in the code.

### LINKED ISSUE #32: [bug] get_cpus may fail sometimes
```
root@E4-C3-2A-90-84-97:/usr/sbin# ./netjson_monitoring '*'
lua: ./netjson_monitoring:226: bad argument #2 to 'tonumber' (integer expected, got string)
stack traceback:
	[C]: in function 'tonumber'
	./netjson_monitoring:226: in function 'get_cpus'
	./netjson_monitoring:260: in main chunk
	[C]: ?
```
Still not sure why this error is occuring, will try to test more until I got some satisfactory results

devkapilbansal:
Fixes #15
Fixes #32

### DIFF
--- a/openwrt-openwisp-monitoring/files/lib/openwisp/interfaces.lua
+++ b/openwrt-openwisp-monitoring/files/lib/openwisp/interfaces.lua
@@ -22,7 +22,9 @@ local specialized_interfaces={
   modemmanager=function(_, interface)
     local modem=uci_cursor.get('network', interface['interface'], 'device')
     local info={}
-    local general=utils.popen('mmcli --output-json -m '..modem):read("*a")
+    local general_file=io.popen('mmcli --output-json -m '..modem)
+    local general=general_file:read("*a")
+    general_file:close()
     if general and pcall(cjson.decode, general) then
       general=cjson.decode(general)
       general=general.modem
@@ -41,7 +43,9 @@ local specialized_interfaces={
       end
     end
 
-    local signal=io.popen('mmcli --output-json -m '..modem..' --signal-get'):read("*a")
+    local signal_file=io.popen('mmcli --output-json -m '..modem..' --signal-get')
+    local signal = signal_file:read("*a")
+    signal_file:close()
     if signal and pcall(cjson.decode, signal) then
       signal=cjson.decode(signal)
       -- only send data if not empty to avoid generating too much traffic
--- a/openwrt-openwisp-monitoring/files/lib/openwisp/interfaces.lua
+++ b/openwrt-openwisp-monitoring/files/lib/openwisp/interfaces.lua
@@ -22,7 +22,7 @@ local specialized_interfaces={
   modemmanager=function(_, interface)
     local modem=uci_cursor.get('network', interface['interface'], 'device')
     local info={}
-    local general=io.popen('mmcli --output-json -m '..modem):read("*a")
+    local general=utils.popen('mmcli --output-json -m '..modem):read("*a")
     if general and pcall(cjson.decode, general) then
       general=cjson.decode(general)
       general=general.modem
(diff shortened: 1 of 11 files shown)


openwrt-openwisp-monitoring/files/lib/openwisp/monitoring_utils.lua
```diff
@@ -54,4 +56,13 @@ function utils.is_excluded(name)
   return name=='lo'
 end
 
+function utils.popen(command)
+  local command_file=io.popen(command)
+  local output_file=assert(io.tmpfile())
+  output_file:write(command_file:read("*a"))
+  command_file:close()
+  output_file:seek('set',0)
+  return output_file
```
nemesisdesign, line 65: shouldn't the file be removed once the output is read? And are you sure using tmp files is really necessary?

Wouldn't it be enough to just do:

```lua
output = commmand_file:read("*a")
command_file:close()
return output
```

We weren't doing this before explicitly for every call.

nemesisdesign (commented):
I did a search in the files of the project and I see two occurrences of `io.popen()` that are not being changed, can you please double check?

See also my comment below.

nemesisdesign (commented):
I did a search in the files of the project and I see two occurrences of `io.popen()` that are not being changed, can you please double check?

See also my comment below.

devkapilbansal:
> I did a search in the files of the project and I see two occurrences of `io.popen()` that are not being changed, can you please double check?
> 
> See also my comment below.

I did it intentionally. The reason being:
- The `utils.popen` was getting cached in some cases which results in breaking the tests. The solution can be to mock define utils.popen too in each file which I don't think is a cleaner solution.
- We need to modify popen only when we are performing operations before closing the file. The only scenario in our case was looping over the output using `:lines()` function.

devkapilbansal:
Hi @nemesisdesign , I updated the code to use split instead as we agreed in the chat.

nemesisdesign (commented):
there's a lot of repetition in this PR, why not create a utility function which does the read, close and returning of the output?

This also ensures that whoever will put their hands on this code will keep reusing our function and avoid messing it up again by forgetting to close a pipe.

nemesisdesign:
> > there's a lot of repetition in this PR, why not create a utility function which does the read, close and returning of the output?
> > This also ensures that whoever will put their hands on this code will keep reusing our function and avoid messing it up again by forgetting to close a pipe.
> 
> I too tried doing something similar. But if I do it in utils function and import it in other sub-modules, then it may not be possible for now to keep them separated from one another. The cache property of `require` is breaking the functionalities in tests

I hope we can find a way to do it, it would be better, I created an issue here: https://github.com/openwisp/openwrt-openwisp-monitoring/issues/61.

openwrt-openwisp-monitoring/files/lib/openwisp/interfaces.lua
```diff
@@ -14,15 +14,23 @@ local ubus=ubus_lib.connect()
...
 
 local interfaces={}
 
+local function popen(command)
+  local command_file = io.popen(command)
+  local output = command_file:read("*a")
+  command_file:close()
+  return output
+end
+
```
nemesisdesign, line 28: duplicating the same function in each file is actually worse Kapil. If duplicating two lines of code is not good, how can duplicating 6 lines of code be better? Please rollback how it was before.

nemesisdesign (changes requested):

nemesisdesign (approved):
Thanks

PR #52, opened 2021-08-10
[fix] Add only existing bridge members in NetJSON output #43

### LINKED ISSUE #43: [bug] Bridge members shown even if they do not exist
In OpenWRT it's possible to define bridge members that do not exist, they'll be added only if they exist and be ignored otherwise.
This is pretty handy in OpenWISP because it's possible to define different bridge members which get added only if the right template is used.

@devkapilbansal can you change the code so that only existing bridge members are collected in the NetJSON monitoring output?

EG:

```json
{
    "name": "br-wifi",
    "bridge_members": [
        "wlan0",
        "wlan1",
        "tap_tcp",
        "tap_tcp_nocryp",
        "tap_udp",
        "tap_udp_nocryp"
    ]
}
```

In the case above, these members do not exist and therefore should not be added:

- "tap_tcp_nocryp",
- "tap_udp",
- "tap_udp_nocryp"

devkapilbansal:
Fixes #43

### DIFF
--- a/openwrt-openwisp-monitoring/files/sbin/netjson-monitoring.lua
+++ b/openwrt-openwisp-monitoring/files/sbin/netjson-monitoring.lua
@@ -128,6 +128,17 @@ for name, interface in pairs(network_status) do
       bridge_members=interface['bridge-members'],
       multicast=interface.multicast
     }
+
+    -- add existing bridge members only
+    if interface['bridge-members'] ~=nil then
+      local bridge_members = {}
+      for _, bridge_member in ipairs(interface['bridge-members']) do
+        if network_status[bridge_member] then
+          table.insert(bridge_members, bridge_member)
+        end
+      end
+      netjson_interface['bridge_members']=bridge_members
+    end
     if wireless_interfaces[name] then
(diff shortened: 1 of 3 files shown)


openwrt-openwisp-monitoring/files/sbin/netjson-monitoring.lua
```diff
@@ -128,6 +128,17 @@ for name, interface in pairs(network_status) do
       bridge_members=interface['bridge-members'],
       multicast=interface.multicast
     }
+
+    -- add existing bridge members only
```
nemesisdesign, line 132: add only interfaces that are present and up as bridge members

openwrt-openwisp-monitoring/tests/test_files/network_data.lua
```diff
@@ -395,7 +395,7 @@ test_data.devices={
   },
   lan1={
   external=false,
-  present=true,
+  present=false,
```
devkapilbansal, line 398: Changed to false to ensure this is not added as bridge_member while testing

nemesisdesign (approved):
This looks good :+1:

PR #56, opened 2021-08-10
[qa] Format Lua Files #51

devkapilbansal:
Closes #51

PR #62, opened 2021-08-13
[enhancement] Make lib modules as non-executable #60

### LINKED ISSUE #60: [enhancement] Do not make openwisp lua script executable
As I see it, the scripts do not have to be executable that are under `/usr/lib/lua/openwisp` on the system.
Please replace INSTALL_BIN with INSTALL_DATA in the openwrt makefile install routine.

I also noticed that the openwsip-agent also has lua scripts under /usr/lib/lua/openwisp.
Does it make sense to use a separate directory?

devkapilbansal:
Closes #60


nemesisdesign (approved):

nemesisdesign (changes requested):
Confilcts appeared after merging https://github.com/openwisp/openwrt-openwisp-monitoring/pull/56, could you please update this PR?

nemesisdesign (approved):

PR #66, opened 2021-09-08
[enhancement] Turn netjson_monitoring into shell tool #59

### LINKED ISSUE #59: [enhancement] Move /sbin/netjson-monitoring.lua into /usr/libexec
From my point of view, this is a script that `only` is used by openwisp-monitoring. 
o avoid polluting the system path, it is usual to move such scripts to /usr/libexec.
I would create a separate subdirectory maned openwisp in this location and would drop this file there.

devkapilbansal:
Closes #59


nemesisdesign (commented):
Looks on the right track to me, @feckert what do you think?

feckert:
I would say that looks quite good.
There is one thing I would like to say, but that's just my opinion.
I think it's always nicer to use a `-` rather than `_` as a separator.
`
netjson_monitoring` -> `netjson-monitoring`

nemesisdesign:
> I would say that looks quite good.
> There is one thing I would like to say, but that's just my opinion.
> I think it's always nicer to use a `-` rather than `_` as a separator.
> ` netjson_monitoring` -> `netjson-monitoring`

I agree, but I am not 100% sure because the name of the openwisp-config agent is openwisp_config, although there are other scripts which use the dash instead of underscores

There's not a strict consistency, but we could fix this in openwisp-config, we could rename the scripts there and leave an alias for backward compatibility. So over time we should get to the point in which we use only one naming style.
What do you think @feckert

nemesisdesign (approved):
LGTM :+1:

nemesisdesign (approved):
Thanks!

nemesisdesign (changes requested):
There are several instances in which netjson_monitoring hasn't been replaced with netson-monitoring.

@devkapilbansal when you do this kind of changes, YOU MUST ALWAYS:

1. use search and replace in your editor through the entire codebase
2. do manual testing before declaring it ready

nemesisdesign (commented):
@devkapilbansal thanks for updating it, let me know when is ready for review & test.

openwrt-openwisp-monitoring/files/netjson-monitoring
```diff
@@ -0,0 +1,51 @@
...
+show_help() {
+	printf "Usage:\n"
+	printf "  netjson-monitoring [OPTIONS...]\n"
+	printf "\n"
+	printf "Use Netjson Monitoring to get current device metrics data.\n"
+	printf "\n"
+	printf "Options:\n"
+	printf "  -h, --help\t\t\t\t: Show this help text\n"
+	printf "  -v, --version\t\t\t\t: Shows version of the agent.\n"
+	printf "  --dump <interfaces to be monitored>\t: Dump device metrics data.\n"
```
nemesisdesign, line 14: I think this point needs more explanation.
It's not really a list of interface to be monitored but a list of interfaces from which traffic data is extracted.

I think it should be something like:

--dump "<interfaces>"

And then below something like:

"<interfaces>" a space separated list of interfaces from which traffic statistics will be collected (tx_bytes, rx_bytes, etc).
Passing an asterisk here will extract traffic statistics from all interfaces.

nemesisdesign (changes requested):
It seems that `netjson-monitoring --dump *` does not work, only `netjson-monitoring --dump "*"` does.

Can you please fix this?

nemesisdesign (approved):

PR #67, opened 2021-09-14
[optimize] Use gzip to compress data #64

### LINKED ISSUE #64: [optimize] Use gzip to compress data
We can reduce the use of memory of more than 70% thanks to gzip.

We should just do the following:

- a basic gzip is provided by busybox, this should be enough, I am not sure whether we should specify this in the dependencies or not, we should double check this
- gzip files after saving them, eg: `gzip $path`, this command also removes the original file
- decompress files before sending the data, eg: `gzip -d $path`

devkapilbansal:
Closes #64

### DIFF
--- a/openwrt-openwisp-monitoring/files/monitoring.agent
+++ b/openwrt-openwisp-monitoring/files/monitoring.agent
@@ -112,7 +112,7 @@ handle_sigusr1() {
 
 send_data() {
 	while true; do
-		for file in "$TMP_DIR"/*.gz; do
+		for file in "$TMP_DIR"/*; do
 			if [ ! -f "$file" ]; then
 				[ "$VERBOSE_MODE" -eq "1" ] && logger -s "No data file found to send." \
 					-p daemon.info
@@ -125,6 +125,8 @@ send_data() {
 			trap "" USR1
 			basefilename=${file##*/}
 			filename=${basefilename%.*}
+			# remove previous saved response if exist
+			[ "$filename" = "response" ] && rm "$file" 2>/dev/null && continue
 			# extra zeroes are added for nanoseconds precision
 			url="$URL&time=$filename.000000"
 			# retry sending data in case of failure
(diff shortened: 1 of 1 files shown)


openwrt-openwisp-monitoring/Makefile
```diff
@@ -19,7 +19,7 @@ define Package/openwisp-monitoring/default
 	CATEGORY:=Administration
 	SECTION:=admin
 	SUBMENU:=openwisp
-	DEPENDS:=+netjson-monitoring $(3)
+	DEPENDS:=+netjson-monitoring +gzip $(3)
```
nemesisdesign, line 22: wait, we don't need the full gzip, we should be able to use the gzip utility provided by busy box, which is shipped by default in OpenWrt:

```
gzip --help
BusyBox v1.33.1 (2021-07-31 21:08:50 UTC) multi-call binary.

Usage: gzip [-cfkdt] [FILE]...

Compress FILEs (or stdin)

	-d	Decompress
	-t	Test file integrity
	-c	Write to stdout
	-f	Force
	-k	Keep input files
``` 

We are space constrained on these devices, we have to be extra careful before adding a new dependency.

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -188,8 +186,11 @@ send_data() {
 					fi
 					break
 				fi
+				#decompress file
```
nemesisdesign, line 189: remember to leave a space after the `#`

nemesisdesign (commented):
Good progress @devkapilbansal, please see my comments below.

openwrt-openwisp-monitoring/Makefile
devkapilbansal: If `gzip` is already available, this shouldn't take any extra space. However, I removed it as I am not sure if the gzip package is heavier than the one provided by busy box

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -213,6 +218,8 @@ send_data() {
 					sleep "$timeout"
 				fi
 			done
+			# compress file again
+			gzip "$filename"
```
nemesisdesign, line 222: is it really needed?

devkapilbansal: I think not now :thinking: 

I added it when we were checking for `.gz` files only. Removing it

nemesisdesign (approved):
Thanks! :+1:

PR #72, opened 2021-09-30
[feature] Try deleting old data if required memory not available #46

### LINKED ISSUE #46: [enhancement] Delete old data if required memory is not available
In #34 we are storing data in temporary storage if specific percentage(required_memory) of total memory is available.

But it may be possible that the server is not reachable for a long time and memory is filled with the json data files.
In that case, previous data files should be deleted to store the new one.

devkapilbansal:
Closes #46


nemesisdesign (changes requested):
Please update this on the latest master.

devkapilbansal:
> It looks good to me. One question @devkapilbansal, how have you been testing this?

For testing this, I change `required_memory` in `openwisp-monitoring` config ensuring files start deleting once the memory gets filled.

PR #76, opened 2021-10-14
[chores] Move logger calls in subroutine #57

### LINKED ISSUE #57: [enhancement] Move logger call into subroutine
The `logger` shell call is used over and over again.
Which makes the code confusing.
I would suggest that you move the call to a subroutine.

https://github.com/openwisp/openwrt-openwisp-monitoring/blob/master/openwrt-openwisp-monitoring/files/monitoring.init
https://github.com/openwisp/openwrt-openwisp-monitoring/blob/master/openwrt-openwisp-monitoring/files/monitoring.agent

devkapilbansal:
Closes #57 

Removed logger calls to simplify the script


nemesifier (commented):
I just merged #72 :rocket:

nemesifier (commented):
I left one comment below.

@feckert would be good if you could double check this as well.

nemesifier (commented):
Hey @devkapilbansal! I think we should revamp this!

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -34,6 +34,27 @@ show_version() {
 
 echoerr() { echo "$@" 1>&2 && exit 1; }
 
+log() {
+	level="$1"
+	type="$2"
+	shift 2
+
+	if [ "$type" = "-v" ] && [ "$VERBOSE_MODE" -ne "1" ]; then
```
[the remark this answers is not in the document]
nemesisdesign: @feckert BTW in recent OpenWrt versions filtering out log levels doesn't work because the logd implementation doesn't support  it.

openwisp-monitoring/files/monitoring.agent
```diff
@@ -50,9 +85,7 @@ check_available_memory() {
 			if [ -f "$file" ]; then
 				rm "$file"
 			else
-				[ "$VERBOSE_MODE" -eq "1" ] \
-					&& logger -s "Not enough memory available, skipping collect data." \
-						-p daemon.warn
+				log -w -v "Not enough memory available, skipping collect data."
```
nemesisdesign, line 88: After reading this code now.. it seems to me that this message shouldn't be logged only in verbose mode because it is not a condition which should happen often and when it happens we most likely want to know. What do you think?

I would change this in a separate commit/PR though, not now. Let's maintain the behavior unaltered in this PR.

openwisp-monitoring/files/monitoring.agent
```diff
@@ -61,14 +94,11 @@ check_available_memory() {
...
-	[ "$VERBOSE_MODE" -eq "1" ] && logger -s "Collecting NetJSON Monitoring data." \
-		-p daemon.info
+	log -i -v "Collecting NetJSON Monitoring data."
 	until [ "$n" -ge 5 ]; do
 		/usr/sbin/netjson-monitoring --dump "$MONITORED_INTERFACES" && break
 
-		if [ "$n" -eq 5 ]; then
-			[ "$VERBOSE_MODE" -eq "1" ] && logger -s "Collecting data failed!" -p daemon.err
-		fi
+		[ "$n" -eq 5 ] && log -e -v "Collecting data failed!."
```
nemesisdesign, line 101:
```suggestion
		[ "$n" -eq 5 ] && log -e -v "Collecting data failed!"
```

openwisp-monitoring/files/monitoring.agent
```diff
@@ -112,17 +141,15 @@ save_data() {
...
 	return 0
 }
 
 send_data() {
 	while true; do
 		for file in "$TMP_DIR"/*; do
 			if [ ! -f "$file" ]; then
-				[ "$VERBOSE_MODE" -eq "1" ] && logger -s "No data file found to send." \
-					-p daemon.info
+				log -i -v "No data file found to send. Checking after $INTERVAL seconds."
```
nemesisdesign, line 152:
```suggestion
				log -i -v "No data file to send. Checking again in $INTERVAL seconds."
```

openwisp-monitoring/files/monitoring.agent
```diff
@@ -149,46 +176,36 @@ send_data() {
...
+					# check if agent was already passing or not to avoid repeating log messages
+					if [ "$FAILING" -eq "0" ]; then
 						FAILING=1
-						logger -s "Data not sent successfully. Response code is \"$response_code\"." \
-							"Error message is $error_message" \
-							"Run with verbose mode to find more." \
-							-t openwisp-monitoring \
-							-p daemon.err
+						[ "$VERBOSE_MODE" -ne "1" ] && log -e -n "Data not sent successfully. Response code is \"$response_code\"." \
+							"Run with verbose mode to find more."
```
nemesisdesign, line 185: There's still repetition here which makes the code hard to read.
I think we need to edit the log function to avoid this.

Ideally we could do something like:

```shell
log -e -n "Data not sent successfully. Response code is \"$response_code\"." \
    -v "Error message is $error_message"
```

The non verbose output should be printed also in verbose mode, while the verbose output is appended to the non verbose output but only when in verbose mode, so in non verbose mode we would have:

```
Data not sent successfully. Response code is "400".`
```

While in verbose mode we would have:

```
Data not sent successfully. Response code is "400". Error message is ERROR_MESSAGE_HERE.
```

openwisp-monitoring/files/monitoring.agent
```diff
@@ -149,46 +176,36 @@ send_data() {
...
-					elif [ "$FAILING" -eq "1" ]; then
-						logger -s "Data sent successfully." \
-							-t openwisp-monitoring \
-							-p daemon.info
+					log -i -v "Data sent successfully."
+					# check if agent was already failing or not to avoid repeating log messages
+					if [ "$FAILING" -eq "1" ]; then
 						FAILING=0
 						rm -f "$RESPONSE_FILE"
+						[ "$VERBOSE_MODE" -ne "1" ] && log -i -n "Data sent successfully."
```
nemesisdesign, line 197: same here: `Data sent successfully.` is repeated twice and there's another unnecessary if here.

nemesifier (changes requested):
Great work @devkapilbansal, I spotted some minor typos and I think there's room for improvement to avoid repetition and simplifying the code further, please see below.

openwisp-monitoring/files/monitoring.agent
devkapilbansal: Hi, let's say that the memory is full. Then after every 5 mins this message will be logged unless the memory is freed once again. I thought that it will add unnecessary noise.

openwisp-monitoring/files/monitoring.agent
devkapilbansal: The condition is here to ensure that the logs are not added unnecessarily when the connection is interrupted due to some issue for a long time.

Discussed here in "Non-verbose mode" https://github.com/openwisp/openwrt-openwisp-monitoring/pull/42#pullrequestreview-708960736

openwisp-monitoring/files/monitoring.agent
devkapilbansal: I added this if so that we don't fill logs when data is transferred normally. In non-verbose mode, it will add this log "Data sent successfully" only once and the line will appear again if the connection is interrupted.
If the variable `FAILING` is set to 1, then only this line will be logged.

nemesifier (commented):
@devkapilbansal the PR history got completely messed up.

nemesifier (approved):
I feel ashamed that this has been sitting here for so much time.
It's  time to move on, I think it's ready to merge.

nemesifier:
@coderabbitai resolve

nemesifier:
@coderabbitai can you do one last full review and let us know if there's any outstanding issues?

nemesifier:
@coderabbitai are you done reviewing or still ongoing?

PR #78, opened 2021-11-18
[change] Remove unneeded ssl package implementation definitions

feckert:
@nemesisdesign @okraits

By cleaning up the package in openwisp-config, this openwisp-monitoring package should also be changed.


nemesisdesign (commented):
Great work @feckert, can you please update the runbuild script too?
https://github.com/openwisp/openwrt-openwisp-monitoring/blob/master/runbuild

This script compiles the package in the CI build.

feckert:
> Great work @feckert, can you please update the runbuild script too? https://github.com/openwisp/openwrt-openwisp-monitoring/blob/master/runbuild
> 
> This script compiles the package in the CI build.

done :+1:

runbuild
```diff
@@ -41,10 +41,7 @@ sed -i '/routing/d' feeds.conf
 # add required packages
 {
 	echo "CONFIG_PACKAGE_netjson-monitoring=y"
-	echo "CONFIG_PACKAGE_openwisp-monitoring-mbedtls=y"
-	echo "CONFIG_PACKAGE_openwisp-monitoring-nossl=y"
-	echo "CONFIG_PACKAGE_openwisp-monitoring-openssl=y"
-	echo "CONFIG_PACKAGE_openwisp-monitoring-wolfssl=y"
+	echo "CONFIG_PACKAGE_openwisp-monitoring=y"
 } >>.config
```
devkapilbansal, line 45: We can remove brackets here :smile:

feckert: We have `echo "CONFIG_PACKAGE_netjson-monitoring=y"` and `echo "CONFIG_PACKAGE_openwisp-monitoring=y"`
So I think this is fine?

nemesisdesign (commented):
This looks good, thanks @feckert.

@devkapilbansal how is the dependency on openwisp-config satisfied during build time?
Is it pulling it from github or is it using the openwrt packages?
Let's keep in mind there's no updated openwisp-config package in the openwrt public feeds yet.

devkapilbansal:
> @devkapilbansal how is the dependency on openwisp-config satisfied during build time? Is it pulling it from github or is it using the openwrt packages? Let's keep in mind there's no updated openwisp-config package in the openwrt public feeds yet.

It is neither pulled from GitHub, not from OpenWRT packages. `runbuild` is generating warning that `openwisp-config` is needed as a dependency for this package which is not present. 
Installing packages will fail on OpenWRT end of openwisp-config is not installed from GitHub. We can define it in runbuild that will also compile an image of openwisp-config 🤔

devkapilbansal:
@feckert can you please add git link of `openwisp-config` in README and runbuild to make it clear that master branch of github should be used

feckert:
From my point of view, this belongs in an extra pull request/commit, since it didn't work before, did it?

I would update the runbuild script to use either the package form the openwrt package feed.
- To do this, however, this [pullrequest](https://github.com/openwrt/packages/pull/17282) must first be merged.
Or we use dependencies from the https://github.com/openwisp/openwisp-config feed.
- However, the package opwenwisp-config from the package feed must first be unistalled, otherwise the build system gets confused or you don't know exactly which package is now used.


feed master (

nemesisdesign:
> From my point of view, this belongs in an extra pull request/commit, since it didn't work before, did it?
> 
> I would update the runbuild script to use either the package from the openwrt package feed.
> 
> * To do this, however, this [pullrequest](https://github.com/openwrt/packages/pull/17282) must first be merged.
> 
> Or we use dependencies from the https://github.com/openwisp/openwisp-config feed.

Let's always use our own repos because we need to have full control on it at all times.

> * However, the package opwenwisp-config from the package feed must first be unistalled, otherwise the build system gets confused or you don't know exactly which package is now used.

The way we use is to simply list the openwisp feed as first, as indicated in the README:
https://github.com/openwisp/openwisp-config#compiling-a-custom-openwrt-image

That makes sure openwisp-config is compiled from our own github repo.

@feckert can you please squash these two commits?
@devkapilbansal then can you please proceed to make sure the build uses our own repo, I thought it was already working that way, this has slipped out of my radar. You can add a separate commit to this PR, we will then merge both, so that once this PR is merged we have both new package and build running fine.

devkapilbansal:
Hi @feckert 
As openwrt/packages#17282 is merged, I think this PR can be updated.

But, the changes may not reflect when we use opkg package manager on device. Do we need to manually install new version of config first before installing `openwisp-monitoring` ? If that's the case, we should add a comment mentioning that or leave this PR on hold until the changes of `openwisp-config` are reflected on OpenWRT devices.

nemesisdesign:
> Hi @feckert As [openwrt/packages#17282](https://github.com/openwrt/packages/pull/17282) is merged, I think this PR can be updated.
> 
> But, the changes may not reflect when we use opkg package manager on device. Do we need to manually install new version of config first before installing `openwisp-monitoring` ? If that's the case, we should add a comment mentioning that or leave this PR on hold until the changes of `openwisp-config` are reflected on OpenWRT devices.

@devkapilbansal it is not really clear to me if this can be merged or not and if not why (what should we do?) can you please clarify?

devkapilbansal:
@nemesisdesign We need to manually install compiled images of openwisp-config if we want to use this change.

<details>
<summary> Installing OpenWISP Config </summary>

```
root@08-00-27-E4-9C-09:~# opkg update
Downloading http://downloads.openwrt.org/releases/19.07.2/targets/x86/64/packages/Packages.gz
Updated list of available packages in /var/opkg-lists/openwrt_core
Downloading http://downloads.openwrt.org/releases/19.07.2/targets/x86/64/packages/Packages.sig
Signature check passed.
Downloading http://downloads.openwrt.org/releases/19.07.2/packages/x86_64/base/Packages.gz
Updated list of available packages in /var/opkg-lists/openwrt_base
Downloading http://downloads.openwrt.org/releases/19.07.2/packages/x86_64/base/Packages.sig
Signature check passed.
Downloading http://downloads.openwrt.org/releases/19.07.2/packages/x86_64/luci/Packages.gz
Updated list of available packages in /var/opkg-lists/openwrt_luci
Downloading http://downloads.openwrt.org/releases/19.07.2/packages/x86_64/luci/Packages.sig
Signature check passed.
Downloading http://downloads.openwrt.org/releases/19.07.2/packages/x86_64/packages/Packages.gz
Updated list of available packages in /var/opkg-lists/openwrt_packages
Downloading http://downloads.openwrt.org/releases/19.07.2/packages/x86_64/packages/Packages.sig
Signature check passed.
Downloading http://downloads.openwrt.org/releases/19.07.2/packages/x86_64/routing/Packages.gz
Updated list of available packages in /var/opkg-lists/openwrt_routing
Downloading http://downloads.openwrt.org/releases/19.07.2/packages/x86_64/routing/Packages.sig
Signature check passed.
Downloading http://downloads.openwrt.org/releases/19.07.2/packages/x86_64/telephony/Packages.gz
Updated list of available packages in /var/opkg-lists/openwrt_telephony
Downloading http://downloads.openwrt.org/releases/19.07.2/packages/x86_64/telephony/Packages.sig
Signature check passed.
root@08-00-27-E4-9C-09:~# opkg install openwisp-config
Unknown package 'openwisp-config'.
Collected errors:
 * opkg_install_cmd: Cannot install package openwisp-config.
root@08-00-27-E4-9C-09:~# opkg install openwisp-config-nossl
Installing openwisp-config-nossl (0.5.0-1) to root...
Downloading http://downloads.openwrt.org/releases/19.07.2/packages/x86_64/packages/openwisp-config-nossl_0.5.0-1_all.ipk
Configuring openwisp-config-nossl.
openwisp: OpenWISP config agent started
```
</details>

feckert:
@devkapilbansal Now I understand what you mean. You are still working with openwrt-19.07!
Then it is clear to me that it does not work. The change was added upstream only in the master branch.
The change is only included in the next stable which will hopefully come in 2022.

The question now is how do we continue here?
I think we need to add a case distinction.
When we make a change to this project, we check against **all** branches that are active used in your project.
But that is difficult!

@nemesisdesign 
Why don't we merge the two projects `openwisp-config` and `openwrt-openwisp-monitoring` into one?
I see different advantages here:
- The source always fits together and can be tested together
- We can use the same release process as openWrt does with different branches (master, openwrt-21.02, openwrt-15.05).
- Source changes could be cherry-picket between branches easily.
- Changes in package management are easier to handle (dependency packagenaming ...). Because not everything has to be taken into account by the branches. It always relates only to this branch

nemesisdesign:
> @nemesisdesign Why don't we merge the two projects `openwisp-config` and `openwrt-openwisp-monitoring` into one? I see different advantages here:
> 
> * The source always fits together and can be tested together
> * We can use the same release process as openWrt does with different branches (master, openwrt-21.02, openwrt-19.07, openwrt-15.05).
> * Source changes could be cherry-picket between branches easily.
> * Changes in package management are easier to handle (dependency packagenaming ...). Because not everything has to be taken into account by the branches. It always relates only to this branch

It may be a good idea, however, it will introduce more overhead for this upcoming release with little short term benefits.
I recommend we focus on getting the new release out as soon as possible, we're about to start pushing this hard in the coming weeks, the new release of OpenWISP will make available a lot of features and bugfixes that many users are looking for and will boost the proejct's adoption and growth.

nemesisdesign (approved):
I am not sure why this has stalled, but it looks good for me, so I am merging.
Thanks a lot @feckert! :+1:

devkapilbansal:
> I am not sure why this has stalled, but it looks good for me, so I am merging. Thanks a lot @feckert! +1

Hi @nemesisdesign I think this PR was on hold until this changes are released for openwisp-config too because it will break with current release of `openwisp-config`

PR #87, opened 2022-04-20
Update Makefile

meferspb:
no compile before in OpenWrt 21.02.2


nemesisdesign:
Unfortunately this is not possible, because it would break newer versions.
Use these instructions to compile also on OpenWrt 19 or earlier directly from the updated openwisp-config source:
https://github.com/openwisp/openwisp-config#compiling-a-custom-openwrt-image

PR #88, opened 2022-04-22
[Fix] Agent exits if device hasn't registered #84

### LINKED ISSUE #84: [bug] Agent exits if device hasn't registered to openwisp controller yet
If the device is not registered yet, the agent exits:

```
openwisp-monitoring: OpenWISP monitoring daemon stopping
openwisp-monitoring: uuid or key are not set, please add these to /etc/config/openwisp
```

This is wrong, because it means users will have to restart the agent manually each time they register a device, which is not what we want.

The agent needs to be smart enough and figure out that if there's a shared secret in the openwisp configuration, it means that openwisp-config has not registered yet but it should not exit like that.

We have to introduce a mechanism which allows the agent to wait for configuration changes and start automatically after the device registers.

I wonder if we can somehow force the openwisp-monitoring agent to automatically restart whenever the config of openwisp-config changes? @feckert do you know if this is possible?

codesankalp:
Fixed by waiting for the uuid and key to be set

Fixes #84

### DIFF
--- a/openwrt-openwisp-monitoring/files/monitoring.agent
+++ b/openwrt-openwisp-monitoring/files/monitoring.agent
@@ -196,6 +196,24 @@ send_data() {
 	done
 }
 
+register() {
+	logger -s "Waiting for device to register" \
+		-t openwisp-monitoring \
+		-p daemon.info
+	# shellcheck disable=SC2002
+	{
+		UUID=$(cmd /etc/config/openwisp | awk '/option uuid / { print $3 }' | tr -d "'")
+		KEY=$(cat /etc/config/openwisp | awk '/option key / { print $3 }' | tr -d "'")
+	}
+	if [ -z "$UUID" ] || [ -z "$KEY" ]; then
+		return 1
+	fi
+	logger -s "Setting uuid and key" \
+		-t openwisp-monitoring \
+		-p daemon.info
+	export UUID KEY
+}
+
 main() {
 	# parse options
 	while [ -n "$1" ]; do
@@ -256,6 +274,7 @@ main() {
 	done
 
 	INTERVAL=${INTERVAL:-300}
+	REGISTRATION_INTERVAL=$((INTERVAL / 10))
 	VERBOSE_MODE=${VERBOSE_MODE:-0}
 	TMP_DIR="/tmp/openwisp/monitoring"
 
@@ -267,6 +286,11 @@ main() {
 		MONITORED_INTERFACES=$(echo "$MONITORED_INTERFACES" | tr -d '"')
 		save_data
 	elif [ "$MODE" = "send" ]; then
+		if [ -z "$UUID" ] || [ -z "$KEY" ]; then
+			until register; do
+				sleep "$REGISTRATION_INTERVAL"
+			done
+		fi
 		VERIFY_SSL=${VERIFY_SSL:-0}
 		RESPONSE_FILE="$TMP_DIR"/response.txt
 		set_url_and_curl && send_data
--- a/openwrt-openwisp-monitoring/files/monitoring.agent
+++ b/openwrt-openwisp-monitoring/files/monitoring.agent
@@ -196,15 +196,15 @@ send_data() {
 	done
 }
 
-register() {
+wait_until_registered() {
+	if [ -n "$UUID" ] && [ -n "$KEY" ]; then
+		return 0
+	fi
 	logger -s "Waiting for device to register" \
 		-t openwisp-monitoring \
 		-p daemon.info
-	# shellcheck disable=SC2002
-	{
-		UUID=$(cmd /etc/config/openwisp | awk '/option uuid / { print $3 }' | tr -d "'")
-		KEY=$(cat /etc/config/openwisp | awk '/option key / { print $3 }' | tr -d "'")
-	}
+	UUID=$(uci get openwisp.http.uuid)
+	KEY=$(uci get openwisp.http.key)
 	if [ -z "$UUID" ] || [ -z "$KEY" ]; then
 		return 1
 	fi
@@ -286,11 +286,9 @@ main() {
 		MONITORED_INTERFACES=$(echo "$MONITORED_INTERFACES" | tr -d '"')
 		save_data
 	elif [ "$MODE" = "send" ]; then
-		if [ -z "$UUID" ] || [ -z "$KEY" ]; then
-			until register; do
-				sleep "$REGISTRATION_INTERVAL"
-			done
-		fi
+		until wait_until_registered; do
+			sleep "$REGISTRATION_INTERVAL"
+		done
 		VERIFY_SSL=${VERIFY_SSL:-0}
 		RESPONSE_FILE="$TMP_DIR"/response.txt
 		set_url_and_curl && send_data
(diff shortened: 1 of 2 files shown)


openwrt-openwisp-monitoring/files/monitoring.init
```diff
@@ -47,10 +47,15 @@ start_service() {
 	fi
 
 	if [ -z "$uuid" ] || [ -z "$key" ]; then
-		logger -s "uuid or key are not set, please add these to /etc/config/openwisp" \
-			-t openwisp-monitoring \
-			-p daemon.err
-		exit 1
+		until [ -n "$uuid" ] && [ -n "$key" ]; do
```
nemesisdesign, line 50: the if before doesn't look necessary anymore

install-dev.sh
```diff
@@ -16,7 +16,7 @@ git clone https://git.openwrt.org/project/libubox.git --depth=1
 git clone https://git.openwrt.org/project/uci.git --depth=1
 { cd uci && cmake . && make install && cd ..; } || { echo 'Installing uci failed!' && exit 1; }
 # install nixio
-luarocks install https://raw.githubusercontent.com/Neopallium/nixio/master/nixio-scm-0.rockspec
+luarocks install https://luarocks.org/nixio-0.3-1.src.rock
```
codesankalp, line 19: The CI is failing due to this line: https://github.com/openwisp/openwrt-openwisp-monitoring/runs/6129392297?check_suite_focus=true#step:6:674

I tried by pointing url to luarocks.org but it shows error in building: https://github.com/openwisp/openwrt-openwisp-monitoring/runs/6131278631?check_suite_focus=true#step:6:690

@nemesisdesign, @devkapilbansal Do you know how to resolve this error?

devkapilbansal: Why you changed the url?

codesankalp: Because the command was failing with previous url, see :point_down: 

image

Command:
```
(base) ➜  openwrt-openwisp-monitoring git:(issues/84-agent-exits) sudo luarocks install https://raw.githubusercontent.com/Neopallium/nixio/master/nixio-scm-0.rockspec
Cloning into 'nixio'...
fatal: remote error: 
  The unauthenticated git protocol on port 9418 is no longer supported.
Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.

Error: Failed cloning git repository.
```

nemesisdesign (changes requested):
During my testing I found our that the init.d process is halted because of this.
If the device cannot register, the init process halting means the device cannot even be rebooted until openwisp-monitoring is running.

Moreover, during my testing I found out that the monitoring agent starts before openwisp-config, so openwisp-config doesn't even start unless I start it manually, if I do that, the monitoring agent finally starts and the init process completes, however, only 1 process starts instead of 2:

```
Wed Apr 27 13:44:36 2022 daemon.err openwisp-monitoring[5128]: missing required --mode option
```

I think the approach is not right, if we introduce this change we will have serious issues.

I think the right solution is to allow the agent to start without uuid and key, and when that happens we assume the device is registering, we should add logic which allows the agent to read the configuration of openwisp_config periodically.

The logic shouldn't be very different from what we have in openwisp-config:
https://github.com/openwisp/openwisp-config/blob/master/openwisp-config/files/openwisp.agent#L833-L843

- derive `REGISTRATION_INTERVAL` from interval / 10 (check every 30 seconds by default)
- write a function that reads the openwisp-config and checks whether uuid and key are present, if present it loads the value in the respective variables and returns with zero exit code, otheriwse it returns with non zero exit code
- the loop should use this function

This logic should be in the agent, not the init.d script.
The init.d script shall not block.
We must ensure that both processes start properly, wait until openwisp-config is registered and then start to work automatically without intervention.

codesankalp:
image

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -267,6 +285,11 @@ main() {
 		MONITORED_INTERFACES=$(echo "$MONITORED_INTERFACES" | tr -d '"')
 		save_data
 	elif [ "$MODE" = "send" ]; then
+		if [ -z "$UUID" ] || [ -z "$KEY" ]; then
+			until register; do
+				sleep "$REGISTRATION_INTERVAL"
+			done
+		fi
```
codesankalp, line 292: Waiting for uuid and key to be set in `send` mode only because collect mode does not depend on uuid and key.

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -196,6 +196,24 @@ send_data() {
...
 
+register() {
+	logger -s "Waiting for device to register" \
+		-t openwisp-monitoring \
+		-p daemon.info
+	# shellcheck disable=SC2002
+	{
+		UUID=$(cmd /etc/config/openwisp | awk '/option uuid / { print $3 }' | tr -d "'")
+		KEY=$(cat /etc/config/openwisp | awk '/option key / { print $3 }' | tr -d "'")
+	}
```
devkapilbansal, line 207: I think it's not good to ignore shellcheck recommendations
```suggestion
	UUID=$(awk '/option uuid / { print $3 }' < /etc/config/openwisp | tr -d "'")
	KEY=$(awk '/option key / { print $3 }' < /etc/config/openwisp | tr -d "'")
```

devkapilbansal (changes requested):

openwrt-openwisp-monitoring/files/monitoring.agent
nemesisdesign: @codesankalp please use `uci get` here and remove the `shellcheck disable=SC2002`, consult the `uci` command docs on openwrt https://openwrt.org/docs/guide-user/base-system/uci or look at how it is used in openwisp-config.

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -267,6 +286,11 @@ main() {
 		MONITORED_INTERFACES=$(echo "$MONITORED_INTERFACES" | tr -d '"')
 		save_data
 	elif [ "$MODE" = "send" ]; then
+		if [ -z "$UUID" ] || [ -z "$KEY" ]; then
```
nemesisdesign, line 289: I see redundancy again here, because the `register` function already checks whether UUID and key are present, I would do the following:

- remove the if here
- rename `register` to `wait_until_registered`
- change the `wait_until_registered` to check whether UUID and KEY are present first, then only if not present look into /etc/config/openwisp and finally if uuid and key are present there export the values or return 1

nemesisdesign (changes requested):
Good progress @codesankalp, let's improve this.

openwrt-openwisp-monitoring/files/monitoring.agent
codesankalp: Updated accordingly.

nemesisdesign (commented):
Looks good, I am going to test it asap.

openwrt-openwisp-monitoring/files/monitoring.agent
```diff
@@ -196,6 +196,24 @@ send_data() {
...
 
+wait_until_registered() {
+	if [ -n "$UUID" ] && [ -n "$KEY" ]; then
+		return 0
+	fi
+	logger -s "Waiting for device to register" \
+		-t openwisp-monitoring \
+		-p daemon.info
+	UUID=$(uci get openwisp.http.uuid 2> /dev/null)
+	KEY=$(uci get openwisp.http.key 2> /dev/null)
```
nemesisdesign, line 207: I redirected standard error to `2> /dev/null` to avoid logging failures of the uci command.

nemesisdesign (approved):

nemesisdesign (approved):

PR #96, opened 2022-05-19
[fix] Fix ULA prefix not available #95 

### LINKED ISSUE #95: [bug] ULA prefix check is bugged
Hello

i tested the openwisp monitoring on openwrt first time today.

Now i get only a error in logread:

`Sat May 14 10:13:31 2022 daemon.err openwisp-monitoring: Data not sent successfully. Response code is "400". Error message is ""Invalid data in \"#/\", validator says:\n\nNone is not of type 'object'"" Run with verbose mode to find more.`

if i run manual the /usr/sbin/netjson-monitoring i get this:

```
root@TestDevice1:/tmp# /usr/sbin/netjson-monitoring --dump *
lua: /usr/lib/lua/openwisp-monitoring/utils.lua:7: attempt to index local 'str' (a nil value)
stack traceback:
	/usr/lib/lua/openwisp-monitoring/utils.lua:7: in function 'split'
	/usr/lib/lua/openwisp-monitoring/interfaces.lua:132: in function 'get_addresses'
	/usr/libexec/netjson-monitoring:165: in main chunk
	[C]: ?
root@TestDevice1:/tmp# /usr/sbin/netjson-monitoring --dump eth0
lua: /usr/lib/lua/openwisp-monitoring/utils.lua:7: attempt to index local 'str' (a nil value)
stack traceback:
	/usr/lib/lua/openwisp-monitoring/utils.lua:7: in function 'split'
	/usr/lib/lua/openwisp-monitoring/interfaces.lua:132: in function 'get_addresses'
	/usr/libexec/netjson-monitoring:165: in main chunk
	[C]: ?
```

any ideas what here fails? With verbose  i don't see more errors, i think the probleme is the lua error.

devkapilbansal:
Closes #95


openwisp-monitoring/tests/test_ula.lua
```diff
@@ -0,0 +1,33 @@
+package.path = package.path .. ";../files/lib/openwisp-monitoring/?.lua" ..
```
nemesisdesign, line 1: please rename it to `test_ula_prefix.lua` and make sure it's added to the `runtests` script otherwise it will not be run during tests.

openwisp-monitoring/files/lib/openwisp-monitoring/interfaces.lua
```diff
@@ -128,8 +128,9 @@ function interfaces.get_addresses(name)
           if utils.starts_with(addr, 'fe80') then
             proto = 'static'
           else
-            local ula = uci_cursor.get('network', 'globals', 'ula_prefix')
-            local ula_prefix = utils.split(ula, '::')[1]
+            local ula_prefix = uci_cursor.get('network', 'globals', 'ula_prefix') or
+                                 ''
```
nemesisdesign, line 132: Is the formatter doing this? It looks really ugly, doesn't it.
Can we find a way to avoid this weirdness and keep it on the line above?

nemesisdesign (changes requested):
Looking good! I see a couple of things that should be improved and a flaw (the test is not added to the runtests script!). See below for more info.

openwisp-monitoring/luaformat.config
```diff
@@ -1,4 +1,4 @@
-column_limit: 86
+column_limit: 88
```
devkapilbansal, line 1: I have to change `column_limit` of the formatter and that introduces changes in other files

nemesisdesign (changes requested):
isn't it possible to just disable the formatter for this particular line?
https://github.com/Koihik/LuaFormatter#disable-formatting-for-a-line-or-block

We cannot change the global configuration every time there's a line which looks off, we need the two packages (this one and openwisp-config) to have consistent configuration and moreover we don't want to pollute the history with unnecessary changes.

nemesisdesign (approved):
Thanks!

PR #99, opened 2022-08-22
[feature] Add random delay before send_data

minhng99:
This can help prevent server overload.

Signed-off-by: Nguyen Quang Minh <hello@minhng99.eu>

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -73,6 +73,7 @@ UCI configuration options must go in ``/etc/config/openwisp-monitoring``.
 
 - ``monitored_interfaces``: interfaces that needs to be monitored, defaults to ``*`` for all interfaces.
 - ``interval``: time after which device data should be sent to server, defaults to ``300``.
+- ``send_delay``: add random wait up to this amount into ``interval`` before send data, defaults to ``10``.
 - ``verbose_mode``: can be enabled (set to ``1``) to ease `debugging <#debugging>`__ in case of issues, defaults to ``0`` (disabled).
 - ``required_memory``: available memory required to save data temporarily, defaults to ``0.05`` (5 percent).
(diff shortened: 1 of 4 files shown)


nemesisdesign (commented):
The agent can recover from failures to send data if the server is unresponsive or if the internet connection of the device is not working.
If the server is overloaded, rather than using these patches, it is advised to scale vertically and/or horizontally the deployment and to optimize the linux system to be able to handle as many connections and open files as possible.
This patch is not a real solution to scaling issues.

What could make sense instead is to add a boot_up delay as we have in openwisp-config: https://github.com/openwisp/openwisp-config/pull/101, which is useful when big power outages happen and many devices boot all at the same time and start sending data all at the same time, the random bootup delay prevents the agents from sending the data all at the same time to the server which is definitely undesirable.

minhng99:
I can't really scale my server right now... but my setup is having an issue whenever I rebooted the server, all the agents send their statistics and failed and just keep trying it again leading to my server getting overloaded when it's just booted up.

Maybe something like an exponential back-off period after sending failed?

nemesisdesign:
Hi @minhng99 

> Maybe something like an exponential back-off period after sending failed could be better?

I created two issues:

- https://github.com/openwisp/openwrt-openwisp-monitoring/issues/107
- https://github.com/openwisp/openwrt-openwisp-monitoring/issues/108

Let's focus our efforts on those, I suggest trying a randomized backoff and a bootup delay on your system and see if that helps.

nemesisdesign:
Superseded by https://github.com/openwisp/openwrt-openwisp-monitoring/pull/111. Thanks for starting the discussion :+1:.

PR #103, opened 2022-10-24
[fix] Discard data on bad request #101

### LINKED ISSUE #101: [bug] Do not retry sending data if HTTP response is 400 bad request
Data should not be sent again if there's a 400 error, the error should be logged as usual and data should be discarded.

devkapilbansal:
Closes #101

### DIFF
--- a/openwisp-monitoring/files/monitoring.agent
+++ b/openwisp-monitoring/files/monitoring.agent
@@ -182,6 +182,13 @@ send_data() {
 					# remove saved data
 					rm -f "$filename"
 					break
+				elif [ "$response_code" = "400" ];then
+					logger -s "Data not sent successfully. Response code is \"$response_code\"." \
+					    "The data being sent is being rejected by the server, discarding data" \
+						-t openwisp-monitoring \
+						-p daemon.err
+					rm -f "$filename"
+					break
 				else
 					timeout=$((timeout * 2))
 					[ "$VERBOSE_MODE" -eq "1" ] && logger -s "Data not sent successfully. Retrying in $timeout seconds" \
--- a/openwisp-monitoring/files/monitoring.agent
+++ b/openwisp-monitoring/files/monitoring.agent
@@ -182,9 +182,9 @@ send_data() {
 					# remove saved data
 					rm -f "$filename"
 					break
-				elif [ "$response_code" = "400" ];then
+				elif [ "$response_code" = "400" ]; then
 					logger -s "Data not sent successfully. Response code is \"$response_code\"." \
-					    "The data being sent is being rejected by the server, discarding data" \
+						"The data being sent is being rejected by the server, discarding data" \
 						-t openwisp-monitoring \
 						-p daemon.err
 					rm -f "$filename"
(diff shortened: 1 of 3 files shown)


openwisp-monitoring/files/monitoring.agent
```diff
@@ -182,6 +184,13 @@ send_data() {
 					# remove saved data
 					rm -f "$filename"
 					break
+				elif [ "$response_code" = "400" ];then
+					logger -s "Data not sent successfully. Response code is \"$response_code\"." \
+					    "Data may be tempered, discarding data" \
+						-t openwisp-monitoring \
+						-p daemon.err
+					rm -f "$filename"
```
devkapilbansal, line 192: Error message is logged in non-verbose mode too since it will be discarded and data points will be missing on controller end

openwisp-monitoring/files/monitoring.agent
```diff
@@ -182,6 +184,13 @@ send_data() {
 					# remove saved data
 					rm -f "$filename"
 					break
+				elif [ "$response_code" = "400" ]; then
+					logger -s "Data not sent successfully. Response code is \"$response_code\"." \
+						"Data may be tempered, discarding data" \
```
nemesisdesign, line 189:
```suggestion
						"The data being sent is being rejected by the server, discarding data" \
```

openwisp-monitoring/files/monitoring.agent
```diff
@@ -182,6 +184,13 @@ send_data() {
 					# remove saved data
 					rm -f "$filename"
 					break
+				elif [ "$response_code" = "400" ]; then
```
nemesisdesign, line 187: Shall we do the same for any 4XX response? Any error response in the range of 400-499 is meant to be a client error and hence is always going to be rejected by the server unless the client modifies the request.

openwisp-monitoring/files/monitoring.agent
```diff
@@ -172,6 +172,8 @@ send_data() {
 					if [ "$VERBOSE_MODE" -eq "1" ]; then
 						logger -s "Data sent successfully." \
 							-p daemon.info
+						# remove response file if request was failing previously
+						[ "$FAILING" -eq "1" ] && FAILING=0 && rm -f "$RESPONSE_FILE"
```
nemesisdesign, line 176: I am not sure I understand this change, why is it necessary?

openwisp-monitoring/files/monitoring.agent
devkapilbansal: I have some doubt regarding that. If we are getting client error, does that mean there is something wrong with the data? It can be due to bad config too.

In such cases, fixing the cause can help sending the data collected previously.

openwisp-monitoring/files/monitoring.agent
devkapilbansal: Reverted the changes

openwisp-monitoring/files/monitoring.agent
nemesisdesign: The other 4XX response type we could get is 404 which I believe will have to be handled as  https://github.com/openwisp/openwisp-config/issues/160, for this I have created https://github.com/openwisp/openwrt-openwisp-monitoring/issues/105, we can leave this code as is for now.

nemesisdesign (approved):
Thanks :+1:

PR #106, opened 2022-11-06
[feature] Allow wireguard proto as type virtual

### LINKED ISSUE #92: [feature] Allow recognizing wireguard interface as of type virtual
If the proto address of an interface is wireguard, we can consider its type to be "virtual".

devkapilbansal:
Fixes #92


openwisp-monitoring/files/sbin/netjson-monitoring.lua
```diff
@@ -164,6 +164,12 @@ for name, interface in pairs(network_status) do
       netjson_interface.statistics = interface.statistics
     end
     local addresses = monitoring.interfaces.get_addresses(name)
+    local virtual_interfaces = {'wireguard'}
+
+    if next(addresses) and
+      monitoring.utils.has_value(virtual_interfaces, addresses[1].proto) then
```
devkapilbansal, line 170: I don't think I need to check all addresses. Only one should be enough.

Correct me if I am wrong 🙃

nemesisdesign (approved):
This is good, thanks! :+1: 

I changed the assertions in the test. That's something I had noticed recently but didn't get time to propose a better alternative, I created an issue for that: https://github.com/openwisp/openwrt-openwisp-monitoring/issues/109.

PR #115, opened 2023-01-02
[tests] Changed JSON string assertions to data structure assertions #109

Aryamanz29:
Closes #109

PR #116, opened 2023-01-02
[fix/change] Don't add the wireless interface information if the chan…

### LINKED ISSUE #113: [bug/change] Do not add wireless interface information if channel is not available
This is an issue I have been observing for a few months.
If a device has some misconfigured wifi interfaces or radios, even though the WiFI interface is not completely brought up and hence not working, the system will report its presence with some important information like channel omitted, eg:

```
wlan1     ESSID: unknown
          Access Point: 46:D1:FA:D2:29:7C
          Mode: Client  Channel: unknown (unknown)
          Center Channel 1: unknown 2: unknown
          Tx-Power: 20 dBm  Link Quality: unknown/70
          Signal: unknown  Noise: -92 dBm
          Bit Rate: unknown
          Encryption: unknown
          Type: nl80211  HW Mode(s): 802.11nacax
          Hardware: 14C3:7915 14C3:7915 [MediaTek MT7915E]
          TX power offset: none
          Frequency offset: none
          Supports VAPs: yes  PHY name: phy1
```

The result is a JSON payload which will be rejected by OpenWISP Monitoring.

For this reason, I think we should change this block of code as follows:

https://github.com/openwisp/openwrt-openwisp-monitoring/blob/8e49082ce7ded3f2ec7e10beedf8c6305f21239f/openwisp-monitoring/files/sbin/netjson-monitoring.lua#L85-L106

We should check whether the channel information is present (`iwinfo.channel` is `nil`) and if it's not present we should not add the `wireless` key to the `netjson_interface` dictionary (lua table) nor look for wifi clients.

I think the whole block shown above could be moved to a separate utility function which is called only after we check that `iwinfo.channel` is not `nil`.

We will need to add a test for this in `test_wifi.lua`.

Aryamanz29:
- Don't add the `wireless interface` information if the channel is not available.

Fixes #113

### DIFF
--- a/openwisp-monitoring/files/sbin/netjson-monitoring.lua
+++ b/openwisp-monitoring/files/sbin/netjson-monitoring.lua
@@ -70,41 +70,49 @@ local host_interfaces = {}
 local dns_servers = {}
 local dns_search = {}
 
+local function get_wireless_netjson_interface(name, iwinfo)
+  local clients = nil
+  local is_mesh = false
+  local netjson_interface = {
+    name = name,
+    type = 'wireless',
+    wireless = {
+      ssid = iwinfo.ssid,
+      mode = monitoring.wifi.iwinfo_modes[iwinfo.mode] or iwinfo.mode,
+      channel = iwinfo.channel,
+      frequency = iwinfo.frequency,
+      tx_power = iwinfo.txpower,
+      signal = iwinfo.signal,
+      noise = iwinfo.noise,
+      country = iwinfo.country
+    }
+  }
+  if iwinfo.mode == 'Ad-Hoc' or iwinfo.mode == 'Mesh Point' then
+    clients = ubus:call('iwinfo', 'assoclist', {device = name}).results
+    is_mesh = true
+  else
+    local hostapd_output = ubus:call('hostapd.' .. name, 'get_clients', {})
+    if hostapd_output then clients = hostapd_output.clients end
+  end
+  if not monitoring.utils.is_table_empty(clients) then
+    netjson_interface.wireless.clients = monitoring.wifi.netjson_clients(clients,
+      is_mesh)
+  end
+  return netjson_interface
+end
+
 -- collect relevant wireless interface stats
 -- (traffic and connected clients)
 for _, radio in pairs(wireless_status) do
   for _, interface in ipairs(radio.interfaces) do
     local name = interface.ifname
-    local is_mesh = false
-    local clients = nil
     if name and not monitoring.utils.is_excluded(name) then
       local iwinfo = ubus:call('iwinfo', 'info', {device = name})
-      local netjson_interface = {
-        name = name,
-        type = 'wireless',
-        wireless = {
-          ssid = iwinfo.ssid,
-          mode = monitoring.wifi.iwinfo_modes[iwinfo.mode] or iwinfo.mode,
-          channel = iwinfo.channel,
-          frequency = iwinfo.frequency,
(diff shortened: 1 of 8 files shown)


nemesisdesign (commented):
This looks good, I will test it in the coming days.

nemesisdesign (changes requested):
After merging #115 some conflicts appeared in the test, could you please fix?

nemesisdesign (changes requested):
Good progress @Aryamanz29, during my testing I found a minor issue, please see below.

openwisp-monitoring/files/sbin/netjson-monitoring.lua
```diff
@@ -70,41 +70,49 @@ local host_interfaces = {}
 local dns_servers = {}
 local dns_search = {}
 
+local function get_wireless_netjson_interface(name, iwinfo)
+  local clients = nil
+  local is_mesh = false
+  local netjson_interface = {
+    name = name,
+    type = 'wireless',
```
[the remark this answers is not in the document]
Aryamanz29: sure

openwisp-monitoring/files/sbin/netjson-monitoring.lua
```diff
@@ -70,43 +70,52 @@ local host_interfaces = {}
...
+      noise = iwinfo.noise,
+      country = iwinfo.country,
+      htmode = htmode
+    }
+  }
+  -- if `iwinfo.channel` is empty then remove
+  -- the `wireless` table from the `netjson_interface`
+  if monitoring.utils.is_empty(iwinfo.channel) then
+    monitoring.utils.remove_table_key(netjson_interface, 'wireless')
+  end
```
nemesisdesign, line 96: This solution is problematic, because the subsequent lines may try to add the clients to `netjson_interface.wireless` in edge cases, therefore in some cases this code may fail.

I think the solution here is to do something like the following (which is what I had originally requested): 

```lua
local netjson_interface = {
  name = name,
  type = 'wireless'
}
-- if channel is missing the WiFi interface is not fully up
-- and hence we avoid including its info because it will be rejected
if monitoring.utils.is_empty(iwinfo.channel) == false then
  wireless = {
    ssid = iwinfo.ssid,
    mode = monitoring.wifi.iwinfo_modes[iwinfo.mode] or iwinfo.mode,
    channel = iwinfo.channel,
    frequency = iwinfo.frequency,
    tx_power = iwinfo.txpower,
    signal = iwinfo.signal,
    noise = iwinfo.noise,
    country = iwinfo.country,
    htmode = htmode
  }
  if iwinfo.mode == 'Ad-Hoc' or iwinfo.mode == 'Mesh Point' then
    clients = ubus:call('iwinfo', 'assoclist', {device = name}).results
    is_mesh = true
  else
    local hostapd_output = ubus:call('hostapd.' .. name, 'get_clients', {})
    if hostapd_output then clients = hostapd_output.clients end
  end
  if not monitoring.utils.is_table_empty(clients) then
    netjson_interface.wireless.clients = monitoring.wifi.netjson_clients(clients,
      is_mesh)
  end
end
return netjson_interface
```

nemesisdesign (changes requested):

openwisp-monitoring/files/sbin/netjson-monitoring.lua
Aryamanz29: I tried this approach initially but it failed many tests. Now I'm able to figure out where it's going wrong, thanks!

nemesisdesign (approved):

PR #131, opened 2024-05-03
[change] Random pause for every 10 successfull sent requests

### LINKED ISSUE #128: [change] Introduce random pause in sending of accumulated data
Introduce random pause in sending of accumulated data at the end of this loop:
https://github.com/openwisp/openwrt-openwisp-monitoring/blob/adf28d536fd61c22ca4f021233e9d8c0bbb49d06/openwisp-monitoring/files/monitoring.agent#L122

The pause should be introduced only every 10 requests, it shall be of a random amount of seconds, comprised between 1 and 5 and it shall be introduced only if the sending was successful.

Dhanus3133:
Closes #128

### DIFF
--- a/openwisp-monitoring/files/monitoring.agent
+++ b/openwisp-monitoring/files/monitoring.agent
@@ -118,6 +118,7 @@ handle_sigusr1() {
 }
 
 send_data() {
+	success_count=0
 	while true; do
 		for file in "$TMP_DIR"/*; do
 			if [ ! -f "$file" ]; then
@@ -166,6 +167,7 @@ send_data() {
 				# send data
 				response_code=$($CURL_COMMAND -H "Content-Type: application/json" -d "$data" "$url")
 				if [ "$response_code" = "200" ]; then
+					success_count=$((success_count + 1))
 					if [ "$VERBOSE_MODE" -eq "1" ]; then
 						logger -s "Data sent successfully." \
 							-p daemon.info
@@ -195,6 +197,11 @@ send_data() {
 			done
 			# retry sending same data again in next cycle
 			[ "$failures" -eq "$MAX_RETRIES" ] && break
+            # pause for a while after every 10 successful request sent
+			if [ $((success_count % 10)) -eq 0 ]; then
+				pause_duration=$(/usr/sbin/openwisp-get-random-number 1 5)
+				sleep "$pause_duration"
+			fi
 		done
 	done
 }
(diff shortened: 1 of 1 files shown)


openwisp-monitoring/files/monitoring.agent
```diff
@@ -195,6 +197,11 @@ send_data() {
 			done
 			# retry sending same data again in next cycle
 			[ "$failures" -eq "$MAX_RETRIES" ] && break
+            # pause for a while after every 10 successful request sent
+			if [ $((success_count % 10)) -eq 0 ]; then
```
pandafy, line 201: Isn't `0 % 10 == 0` ? Why not just use `[ $success_rate -ge 10 ]` which is more readable?

pandafy (changes requested):

openwisp-monitoring/files/monitoring.agent
```diff
@@ -195,6 +197,11 @@ send_data() {
 			done
 			# retry sending same data again in next cycle
 			[ "$failures" -eq "$MAX_RETRIES" ] && break
+            # pause for a while after every 10 successful request sent
```
nemesifier, line 200: watch out from indentation mismatches like this

openwisp-monitoring/files/monitoring.agent
```diff
@@ -118,6 +118,7 @@ handle_sigusr1() {
 }
 
 send_data() {
+	success_count=0
```
nemesifier, line 121: If we start from zero, the pause will not be executed after 10 requests, but 11, right? So If we want the implementation to follow the spec and if we want the comment below to be technically correct, I think we should start from 1 here.

openwisp-monitoring/files/monitoring.agent
Dhanus3133: `-ge` makes it run whenever the `success_rate` is more than or equal to 10. But as per the issue, it was meant for every 10 requests. Was I right on this one?

openwisp-monitoring/files/monitoring.agent
Dhanus3133: Sorry MB, formatting issue will fix it

openwisp-monitoring/files/monitoring.agent
Dhanus3133: MB you're right. Will update it!

openwisp-monitoring/files/monitoring.agent
nemesifier: The requirement is to pause every 10 successful requests.
I don't see any `success_rate` variable in this diff, @pandafy where is `success_rate` coming from?
I only see `success_count` here.

openwisp-monitoring/files/monitoring.agent
```diff
@@ -118,6 +118,7 @@ handle_sigusr1() {
 }
 
 send_data() {
+	success=1
```
nemesifier, line 121: After having reviewed the code I think you were right in setting this to zero, my calculation was incorrect.

nemesifier (approved):

nemesifier (approved):

Issue #134, opened 2024-06-23
[enhancement] openwisp monitoring.agent cacert and capath options for curl like on openwisp.agent


SLSC1907:
Hi there.

I'm running openwisp_config with the cacert option.
When i was trying to configure openwisp-monitoring I noticed that it has the verify_ssl option, but doesn't  have the cacert or capath options, like openwisp_config does.

PR #135, opened 2024-07-03
[enhancement] Add curl cacert capath options #134

SLSC1907:
Addresses #134

Update monitoring.init and monitoring.agent for curl cacert capath options from openwisp-config configuration.

### DIFF
--- a/openwisp-monitoring/files/monitoring.agent
+++ b/openwisp-monitoring/files/monitoring.agent
@@ -17,8 +17,8 @@ show_help() {
 	printf "  --uuid <uuid>\t\t\t\t: UUID of the device.\n"
 	printf "  --key <key>\t\t\t\t: Key for the device.\n"
 	printf "  --verify_ssl <0, 1>\t\t\t: Whether SSL Authentication should be enabled or not.\n"
-        printf "  --cacert <file>\t\t\t: Value passed to curl --cacert argument.\n"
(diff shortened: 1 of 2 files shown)


nemesifier (changes requested):
Looks good @SLSC1907, please add these options to the [Configuration Options section in the README](https://github.com/openwisp/openwrt-openwisp-monitoring?tab=readme-ov-file#configuration-options).

SLSC1907:
Thanks for the feedback.
I didn't include it as the documentation and configuration is set from the openwrt-openwisp-config package.

i.e. These UCI options aren't read from `/etc/config/openwisp-monitoring` but are pulling the options from /etc/config/openwisp just like the existing verify_ssl option.

As it exists, if you use the cacert and capath options with verify_ssl in openwisp-config, the monitoring package will fail to upload / verify the certificate.

The options are documented however in the monitoring.agent printed help output, like the verify_ssl option.

Would you like all the Openwisp general config items documented in the readme for openwisp-monitoring?

My plan is to add a note underneath regarding the application also pulls configuration options from /etc/config/openwisp. I can write a separate documentation update too? Not sure how you'd like to proceed.

nemesifier (approved):

nemesifier (approved):
@SLSC1907 you're right, documenting these options is not needed, I had missed this detail in my previous review, thanks for pointing it out. I fixed some issues with whitespace (we use tabs here instead of spaces), not it can be merged. Thanks for contributing! :pray:

PR #152, opened 2025-04-01
Checksum1.1

### LINKED ISSUE #74: [ci] Upload checksum value for packages
Since we have already added a functionality to upload packages directly to downloads.openwisp.io, it will be beneficial to also upload checksums for built packages.

This way users will be able to verify the authenticity of downloaded packages.

alookik27:
## Checklist

- [x] I have read the [OpenWISP Contributing Guidelines](http://openwisp.io/docs/developer/contributing.html).
- [x] I have manually tested the changes proposed in this pull request.

## Reference to Existing Issue

Closes #74 .

## Description of Changes

Fortunately i have just edit the ci.yml file ,and add the SHA-256 checksum before uploading the packages to downloads.openwisp.io .

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,19 +100,27 @@ jobs:
         with:
           project_id: ${{ secrets.GCS_PROJECT_ID }}
           credentials_json: ${{ secrets.GCS_DOWNLOADS_SERVICE_ACCOUNT_JSON }}
-
       - name: Google Cloud Setup
         if: ${{ github.event_name=='push' }}
         uses: google-github-actions/setup-gcloud@v2
         with:
           project_id: ${{ secrets.GCS_PROJECT_ID }}
+  
+      - name: Generate checksums for compiled packages
+        if: ${{ github.event_name=='push' }}
+        run: |
+          find ${{ env.DOWNLOADS_DIR }}/${{ env.START_TIME }}/openwisp_monitoring -type f -name "*.ipk" -exec sha256sum {} \; > ${{ env.DOWNLOADS_DIR }}/${{ env.START_TIME }}/openwisp_monitoring/sha256sums.txt
+  
+      - name: Define Storage Paths
+        run: |
+          echo "SRC_URL=${{ env.DOWNLOADS_DIR }}/${{ env.START_TIME }}/openwisp_monitoring" >> $GITHUB_ENV
+          echo "DST_URL=gs://${{ secrets.GCS_DOWNLOADS_BUCKET_NAME }}/openwisp-monitoring/${{ env.START_TIME }}-${{ env.COMMIT_SHA }}" >> $GITHUB_ENV
+          echo "LATEST_URL=gs://${{ secrets.GCS_DOWNLOADS_BUCKET_NAME }}/openwisp-monitoring/latest" >> $GITHUB_ENV
 
       - name: Upload compiled packages to downloads.openwisp.io
         if: ${{ github.event_name=='push' }}
         run: |
-          gsutil -m rsync -r ${{ env.SRC_URL }} ${{ env.DST_URL }}
-          gsutil -m rsync -r -d ${{ env.SRC_URL }} ${{ env.LATEST_URL }}
-        env:
-          SRC_URL: ${{ env.DOWNLOADS_DIR }}/${{ env.START_TIME }}/openwisp_monitoring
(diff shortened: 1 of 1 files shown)


nemesifier (changes requested):
Can you please show a visual result of how the user can find out about the checksum?
And show the generated output?

nemesifier (changes requested):
As discussed during the weekly call, the compilation steps generates a file which contains the checksum.
I recommend compiling locally, find the file which contains the checksum and publish it on Google Cloud as "Packages.sha256.checksum".

.github/workflows/ci.yml
```diff
@@ -100,19 +100,27 @@ jobs:
...
+      - name: Generate checksums for compiled packages
+        if: ${{ github.event_name=='push' }}
+        run: |
+          find ${{ env.DOWNLOADS_DIR }}/${{ env.START_TIME }}/openwisp_monitoring -type f -name "*.ipk" -exec sha256sum {} \; > ${{ env.DOWNLOADS_DIR }}/${{ env.START_TIME }}/openwisp_monitoring/sha256sums.txt
+  
+      - name: Define Storage Paths
+        run: |
+          echo "SRC_URL=${{ env.DOWNLOADS_DIR }}/${{ env.START_TIME }}/openwisp_monitoring" >> $GITHUB_ENV
+          echo "DST_URL=gs://${{ secrets.GCS_DOWNLOADS_BUCKET_NAME }}/openwisp-monitoring/${{ env.START_TIME }}-${{ env.COMMIT_SHA }}" >> $GITHUB_ENV
+          echo "LATEST_URL=gs://${{ secrets.GCS_DOWNLOADS_BUCKET_NAME }}/openwisp-monitoring/latest" >> $GITHUB_ENV
```
devkapilbansal, line 118: Why this change was required? We are already defining `env` in the next command where they are needed

.github/workflows/ci.yml
```diff
@@ -100,19 +100,27 @@ jobs:
...
-          gsutil -m rsync -r ${{ env.SRC_URL }} ${{ env.DST_URL }}
-          gsutil -m rsync -r -d ${{ env.SRC_URL }} ${{ env.LATEST_URL }}
-        env:
-          SRC_URL: ${{ env.DOWNLOADS_DIR }}/${{ env.START_TIME }}/openwisp_monitoring
-          DST_URL: gs://${{ secrets.GCS_DOWNLOADS_BUCKET_NAME }}/openwisp-monitoring/${{ env.START_TIME }}-${{ env.COMMIT_SHA }}
-          LATEST_URL: gs://${{ secrets.GCS_DOWNLOADS_BUCKET_NAME }}/openwisp-monitoring/latest
+          gsutil -m rsync -r $SRC_URL $DST_URL
+          gsutil -m rsync -r -d $SRC_URL $LATEST_URL
+          gsutil cp $SRC_URL/sha256sums.txt $DST_URL/
+          gsutil cp $SRC_URL/sha256sums.txt $LATEST_URL/
```
devkapilbansal, line 126: If we are already syncing the source with destination, then why we need to add checksum files ourselves? Shouldn't they will be added automatically in the bucket if they are available?

devkapilbansal (changes requested):

devkapilbansal:
> Can you please show a visual result of how the user can find out about the checksum? And show the generated output?

+1

devkapilbansal:
@alookik27 Ping!

.github/workflows/ci.yml
alookik27: Sorry to answer you so late ,i have reason besides that 
the reason for defining the env variable "Define Storage Paths" is for readability ,reuse because the paths are bit long and used in many commands i think it will be simple for maintainability

i edited the file while learning yaml so i searched and found out that adding the env variable can be good for reuse
if these change are bad then i would remove them and edit the file .

.github/workflows/ci.yml
alookik27: for the safety reason if anything happens than the verification can be smoothly done by checksum file.

.github/workflows/ci.yml
devkapilbansal: It would have been good if these env variables were used anywhere else. These variables are specific for Google Cloud only and therefore added there.

I don't think moving them upwards solves any for our use case.

.github/workflows/ci.yml
devkapilbansal: Do you mean that in any case the packages are not uploaded online, then we can check it via Checksum? I am not very much convinced.

What do you think @nemesifier ?

devkapilbansal (changes requested):

